PR #1, opened 2022-12-14
Tests: Adds lzip decoder tests

JiaT75:
## Pull request checklist

- [X] Docs have been reviewed and added / updated if needed (for bug fixes / features)

## Pull request type

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Other (please describe): Test for recently added feature

This PR does not alter behavior

Related Issue URL: N/A

## What is the new behavior?

N/A

- [ ] Yes
- [X] No

### DIFF
--- a/tests/test_lzip_decoder.c
+++ b/tests/test_lzip_decoder.c
@@ -1,0 +1,471 @@
+///////////////////////////////////////////////////////////////////////////////
+//
+/// \file       test_lzip_decoder.c
+/// \brief      Tests decoding lzip data
+//
+//  Author:     Jia Tan
+//
+//  This file has been put into the public domain.
+//  You can do whatever you want with this file.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+#include "tests.h"
+
+#ifdef HAVE_LZIP_DECODER
+
+// Memlimit large enough to pass all of the test files
+#define MEMLIMIT (1U << 20)
+#define DECODE_CHUNK_SIZE 1024
+
+
+// Avoiding using data buffers so we don't have to store the data buffers
+// as large hex strings. Instead, store the CRC32 value of the expected data.
+// CRC32 value of "Hello\nWorld\n"
+static const uint32_t hello_world_crc = 0x15A2A343;
+
+// CRC32 value of "Trailing garbage\n"
+static const uint32_t trailing_garbage_crc = 0x87081A60;
+
+
+// Helper function to decode a good file with no flags and plenty high memlimit
+static void
+basic_lzip_decode(const char *src, const uint32_t expected_crc) {
+	size_t file_size;
+	uint8_t *data = tuktest_file_from_srcdir(src, &file_size);
+	uint32_t checksum = 0;
+
+	lzma_stream strm = LZMA_STREAM_INIT;
+	assert_lzma_ret(lzma_lzip_decoder(&strm, MEMLIMIT, 0), LZMA_OK);
+
+	uint8_t *output_buffer = tuktest_malloc(DECODE_CHUNK_SIZE);
+
+	strm.next_in = data;
+	strm.next_out = output_buffer;
+	strm.avail_out = DECODE_CHUNK_SIZE;
+
+	// Feed 1 byte at a time to the decoder to look for any bugs
+	// when switching between decoding sequences
+	lzma_ret ret = LZMA_OK;
+	while (ret == LZMA_OK) {
+		strm.avail_in = 1;
+		ret = lzma_code(&strm, LZMA_RUN);
+		if (strm.avail_out == 0) {
+			checksum = lzma_crc32(output_buffer,
+					strm.next_out - output_buffer,
+					checksum);
+			// No need to free output_buffer because it will
+			// automatically be freed at the end of the test by
+			// tuktest.
+			output_buffer = tuktest_malloc(DECODE_CHUNK_SIZE);
+			strm.next_out = output_buffer;
+			strm.avail_out = DECODE_CHUNK_SIZE;
+		}
+	}
+
+	assert_lzma_ret(ret, LZMA_STREAM_END);
+	assert_uint_eq(strm.total_in, file_size);
+
+	checksum = lzma_crc32(output_buffer, strm.next_out - output_buffer,
+			checksum);
+	assert_uint_eq(checksum, expected_crc);
+
+	lzma_end(&strm);
+}
+
+
+static void
+test_options(void)
+{
+	// Test NULL stream
+	assert_lzma_ret(lzma_lzip_decoder(NULL, MEMLIMIT, 0),
+			LZMA_PROG_ERROR);
+
+	// Test invalid flags
+	lzma_stream strm = LZMA_STREAM_INIT;
+	assert_lzma_ret(lzma_lzip_decoder(&strm, MEMLIMIT, UINT32_MAX),
+			LZMA_OPTIONS_ERROR);
+	// Memlimit tests are done elsewhere
+}
+
+
+static void
+test_v0_decode(void) {
+	// This tests if liblzma can decode lzip version 0 files.
+	// lzip 1.17 and older can decompress this, but lzip 1.18
+	// and newer can no longer decode these files.
+	basic_lzip_decode("files/good-1-v0.lz", hello_world_crc);
+}
+
+
+static void
+test_v1_decode(void) {
+	// This tests decoding a basic lzip v1 file
+	basic_lzip_decode("files/good-1-v1.lz", hello_world_crc);
+}
+
+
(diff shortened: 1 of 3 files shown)


tests/test_lzip_decoder.c
```diff
@@ -0,0 +1,454 @@
...
+//  You can do whatever you want with this file.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+#include "tests.h"
+
+#ifdef HAVE_LZIP_DECODER
+
+// Memlimit large enough to pass all of the test files
+#define MEMLIMIT 1 << 20
```
Larhzu, line 18: Parenthesis are needed, unsigned value might be nicer but doesn't matter. #define MEMLIMIT (1U << 20)

tests/test_lzip_decoder.c
```diff
@@ -0,0 +1,454 @@
...
+#define DECODE_CHUNK_SIZE 1024
+
+
+// Helper function to decode a good file with no flags and plenty
+// high memlimit
+static void
+basic_lzip_decode(const char* src, const uint8_t* expected_result,
```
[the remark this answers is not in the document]
JiaT75: I know I'm quite inconsistent with this. Thanks for pointing it out (again)

tests/test_lzip_decoder.c
```diff
@@ -0,0 +1,454 @@
...
+	assert_lzma_ret(lzma_lzip_decoder(&strm, MEMLIMIT, UINT32_MAX),
+			LZMA_OPTIONS_ERROR);
+	// Memlimit tests are done elsewhere
+}
+
+
+static void
+test_v0_decode(void) {
+	// This tests if liblzma can decode lzip version 0 files.
+	// lzip 1.17 and older can decomperss this, but lzip 1.18
```
Larhzu, line 75: "decompress" typo

tests/test_lzip_decoder.c
```diff
@@ -0,0 +1,454 @@
...
+	// and newer can no longer decode these files.
+	const char expected_result[] = "Hello\nWorld!\n";
```
Larhzu, line 77: ASCII strings aren't portable to EBCDIC systems and thus cannot be used when looking for byte-exact results.

tests/test_lzip_decoder.c
```diff
@@ -0,0 +1,454 @@
...
+	basic_lzip_decode("files/good-1-v1.lz",
```
Larhzu, line 78: The test files are tested by test_files.sh already. Replacing test_files.sh by something else is one question but duplicating the test here doesn't seem useful.

tests/test_lzip_decoder.c
```diff
@@ -0,0 +1,454 @@
...
+	strm.avail_in = file_size;
+	strm.avail_out = expected_size;
+
+	assert_lzma_ret(lzma_code(&strm, LZMA_FINISH), LZMA_STREAM_END);
+	assert_uint(strm.total_in, <, file_size);
+	assert_uint_eq(strm.total_out, expected_size);
+	assert_array_eq(output_buffer, expected_result,
+			strm.total_out);
+
+	// Trailing data should be readable from strm.next_in
```
Larhzu, line 119: This test is useful as it's not done by test_files.sh.

tests/test_lzip_decoder.c
JiaT75: Its useful if we break something in the lzip decoder that causes this to fail. This will tell us the lzma_ret value, but the test_files.sh won't. Also, this checks the actual contents of the file match, but test_files.sh just check if the decoder returns success or not.

JiaT75: I made a typo anyway. It should be:
	basic_lzip_decode("files/good-1-v0.lz",
                        (const uint8_t *) expected_result,
			sizeof(expected_result) - 1);

tests/test_lzip_decoder.c
```diff
@@ -0,0 +1,454 @@
...
+
+static void
+test_invalid_member_size(void) {
+	decode_expect_error("files/bad-1-v1-member-size.lz",
+			LZMA_DATA_ERROR);
+}
+
+
+static void
+test_invalid_memlimit(void) {
```
Larhzu, line 393: Perhaps this test is useful but for portability it could be simplified so that it doesn't check the output content or even size, the CRC32 will catch some errors anyway. That is, just focus on the memlimit testing only.

JiaT75: I think there is value in showing that if we have too low memlimit, we can up it and continue decoding. This seems like a common use case that we need to assert works

tests/test_lzip_decoder.c
Larhzu: Content checking is done by test_files.sh thanks to CRC32, that part is good enough. The return value isn't checked by test_files.sh, that is true. I'm not sure how important it is but it's true it's not much code to check it here to be sure.

tests/test_lzip_decoder.c
Larhzu: That's exactly what I tried to say.

tests/test_lzip_decoder.c
```diff
@@ -0,0 +1,454 @@
...
+
+
+// Avoiding using the ASCII string because of potential EBCDIC system
+// compatibility. The string is: "Hello\nWorld!\n"
+static const char hello_world[] = {'H', 'e', 'l', 'l', 'o', '\n',
+			'W', 'o', 'r', 'l', 'd', '!', '\n', 0};
+// The string is "Trailing garbage\n"
+static const char trailing_garbage[] = {'T', 'r', 'a', 'i', 'l', 'i',
+		'n', 'g', ' ', 'g', 'a', 'r', 'b', 'a', 'g', 'e', '\n', 0};
+
```
Larhzu, line 29: Putting ASCII chars as separate chars instead of as a string doesn't change anything. On EBCDIC systems the source files are mass-converted from ASCII to EBCDIC before compilation and thus the characters map to different byte values. I think it's enough to check for trailing garbage string and that can be done by comparing CRC32 value, that is, compare the length of the garbage and use lzma_crc32() and compare against known hardcoded value (no need for a #define for the hardcoded value). It's ugly but then it's portable and there's no need for a ASCII string in hex representation.

tests/test_lzip_decoder.c
```diff
@@ -0,0 +1,454 @@
...
+	assert_lzma_ret(lzma_lzip_decoder(&strm, MEMLIMIT, UINT32_MAX),
+			LZMA_OPTIONS_ERROR);
+	// Memlimit tests are done elsewhere
+}
+
+
+static void
+test_v0_decode(void) {
+	// This tests if liblzma can decode lzip version 0 files.
+	// lzip 1.17 and older can decompress this, but lzip 1.18
```
Larhzu, line 84: I wonder if it's not useful to repeat the version numbers of the other tool here as things might change and they aren't relevant for liblzma testing. They are described in the tests/files/README in detail already.

tests/test_lzip_decoder.c
```diff
@@ -0,0 +1,454 @@
...
+	// and newer can no longer decode these files.
+	basic_lzip_decode("files/good-1-v0.lz",
+			(const uint8_t *) hello_world,
```
Larhzu, line 87: Style: No space after cast. (const char *)hello_world,

tests/test_lzip_decoder.c
JiaT75: Maybe this is too philosophical about the purpose of the tests, but using the checksum as the content verification assumes that the checksum was properly checked. The more paranoid approach is to also verify the checksum value

tests/test_lzip_decoder.c
JiaT75: I suppose the simple tedious solution is to convert the ASCII characters to hex and hardcode those, but I like hardcoding the CRC better since we shouldn't ever regenerate the test file with a new checksum type. I will work on implementing that

tests/test_lzip_decoder.c
Larhzu: I don't disagree, it's just that I feel it's not that important for these specific tests to be extremely paranoid. Using lzma_crc32() here after decompression to check the uncompressed data might work too? I mostly would like to avoid the need to have the full uncompressed file content in the C code but since that is fairly short I suppose it's OK too, it just has to be in hex so that it survives conversion to EBCDIC. So if you prefer that it's OK.

tests/test_lzip_decoder.c
JiaT75: I would rather check the CRC on the output to make this a more self contained test. What if the memlimit code somehow disables CRC checking and corrupts the output? Reading the lzip_decoder.c, its obvious this isn't the case, but the test doesn't know that. Its very few lines of code to check the output and makes the test more complete

PR #30, opened 2023-02-04
Include Doxygen html files in release

JiaT75:
## Pull request checklist

## Pull request type

 

- [ ] Bugfix
- [ ] Feature
- [X] Build related changes
- [X] Documentation content changes

## What is the new behavior?

- The Doxygen html docs are now included in the release tarball
- If only the liblzma API function documentation is generated, the docs are created in /doc/liblzma
- If all files documentation is generated, the docs are created in /doc/html

- [ ] Yes
- [X] No

## Other information
Putting the docs in liblzma folder may be unnecessary. Other distributions seems to place their html documentation files in a directory called "html" in the doc directory. If we just want to include it in the "html" directory in /doc, then the first and third commit can be omitted.


Makefile.am
```diff
@@ -100,6 +101,8 @@ dist-hook:
 				> "$$dest/txt/$$BASE.txt"; \
 		done; \
 	fi
+	doxygen Doxyfile
+	cp -r "doc/$(doxygen_html_dir)" "$(distdir)/doc/"
```
JiaT75, line 105: Should be -R since -r is not POSIX

JiaT75:
This includes the doxygen .html files in the release, but does not install them. I will need to rework this to actually install the files, which seems more useful than just including them.

JiaT75:
Now I have the HTML files being installed. They are currently installed to `$(docdir)/liblzma`, for liblzma headers only, or `$(docdir)/html` for full documentation. So, by default, it will install to `/share/doc/xz/liblzma`, which I am not sure if I like.

I don't know of a good clean way to install to `/share/doc/liblzma` (or  `/share/doc/liblzma/html` if that makes more sense) since `$(docdir)` is `/share/doc/xz`. We can always install to `$(docdir)/../liblzma`, but that is hacky. If there is a good clean way to do this, I might be interested to try it.

If we don't like how it is now, the easiest alternative is to always install to `/share/doc/xz/html` and not care about the distinction between liblzma header only documentation and all source files documentation.

Issue #18, opened 2023-01-12
[Feature Request]: Create Windows CI Support


JiaT75:
### Describe the Feature

Both MinGw and CMake Windows builds should be tested.

### Expected Complications

Downloading the correct build environment dependencies could be tricky. GitHub might have some reusable components to help with this.

### Will I try to implement this new feature?

Yes

arixmkii:
@JiaT75 Could you specify which flavors of msys2 would you like to support?

There are different combinations of arch, compiler, c standard library: https://www.msys2.org/docs/environments/

I had recent experience of setting GitHub workflows for Windows builds and msys2, I might be able to help.

JiaT75:
@arixmkii Eventually, I would like to support the MSYS, UCRT64, and MINGW64 flavors in CI runners. Along with MSVC CMake of course. The other ones either don't run on x86-64 (which is what the GitHub-hosted runners support), or are the same but just use clang. The clang build is already tested by the MacOS runner, so its probably not necessary to test it on Windows. 

I don't develop or use xz or liblzma on Windows, but we have a few users who do. My note about MinGw and CMake was more about testing the autotool and CMake builds on Windows. So, I think using msys2 makes sense since it is likely many of our users do so.

To start, I think just picking one environment would be great. I feel the most useful one would be UCRT64 if we also have CMake testing msvcrt. Since we already have a lot of runners and configurations, maybe the other environments could be in a separate Workflow that we run manually pre-release or if we are addressing a Windows specific bug.

Thanks for offering to help with this! I was planning on working on this soon, but I kept delaying it to work on other things. Your help would be greatly appreciated if you have the time to do it :)

arixmkii:
@JiaT75 I did something. At least it runs full autotools set on Windows host. I don't yet understand CMake part and also VisualStudio parts, but might be able to take another look some time later. Any additional context, what you think could help is highly appreciated.

JiaT75:
>  Any additional context, what you think could help is highly appreciated.

I am not sure how experienced you are with CMake, but we shouldn't need anything too complicated. I haven't fully thought through it, but here are a few ideas I have had so far:
- The GitHub Windows runner comes with CMake by default. You will not need to install it manually (https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md)
[...]

PR #34, opened 2023-02-12
Enable CI on Windows hosts for autotools

arixmkii:
## Pull request checklist

- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been reviewed and added / updated if needed (for bug fixes / features)

## Pull request type

 

- [ ] Bugfix
- [ ] Feature
- [X] Build related changes

No Windows CI at all.

Related Issue URL: https://github.com/tukaani-project/xz/issues/18

## What is the new behavior?

- autotools build set for Windows added
- build is using msys2 with UCRT64 toolchain

- [ ] Yes
- [X] No

## Other information

CMake build part is not provided as part of this PR.

Additional changes:
* bumped checkout action to v3 (not really critical, but I did while trying to understand build failure)
* fixed "false positive" uninitialized warning with empty in stream_encoder.c
* fixed missing include in memcmplen.h, because this include is not provided if --disable-threads is requested for build

Windows build run is very slow (comparing to others), but I don't think it could be somehow improved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Install Dependencies
         run: sudo apt-get install -y autoconf automake build-essential po4a autopoint gcc-multilib
 
@@ -97,7 +97,7 @@ jobs:
   Linux-CMake:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Install Dependencies
         run: sudo apt-get install -y build-essential cmake
       - name: Build
@@ -115,7 +115,7 @@ jobs:
     runs-on: macos-latest
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Install Dependencies
         run: brew install autoconf automake libtool po4a
 
@@ -169,7 +169,7 @@ jobs:
   MacOS-CMake:
     runs-on: macos-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Install Dependencies
         run: brew install cmake
       - name: Build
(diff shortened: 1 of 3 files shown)


arixmkii:
This is how it looked in my fork.

image

src/liblzma/common/memcmplen.h
```diff
@@ -19,6 +19,9 @@
 #	include <immintrin.h>
 #endif
 
+#if defined(_M_X64) // MSVC or Intel C compiler on Windows
```
JiaT75, line 22: Its interesting that this was only highlighted when threads are disabled. It must have something to do with the `_WIN32_WINNT 0x0600` in `mythread.h`. Either way, the proper include was missing, so thank you for addressing this.

On Windows, <intrin.h> also includes <immintrin.h>, so including both is redundant. Instead, these can be combined into something like this:

```
#if defined(TUKLIB_FAST_UNALIGNED_ACCESS) \
		&& (defined(__INTEL_COMPILER) && defined(_M_X64)) \
			|| (defined(_MSC_VER) && defined(_M_X64)))
#        define <intrin.h>
#else if HAVE_IMMINTRIN_H
#        define <immintrin.h>
#endif
```

Also, please include this as a separate commit from the CI changes. It was discovered because of the CI changes, but the fix is unrelated.

src/liblzma/common/stream_encoder_mt.c
```diff
@@ -743,7 +743,7 @@ stream_encode_mt(void *coder_ptr, const lzma_allocator *allocator,
 
 		// These are for wait_for_work().
 		bool has_blocked = false;
-		mythread_condtime wait_abs;
+		mythread_condtime wait_abs = {};
```
JiaT75, line 746: Similar to the other comment, include this as a separate commit from the CI changes. 

ts funny that it flags this as uninitialized, but not on line 978 in same file. Maybe since its in a switch-case? No change needed to the other use of `mythread_condtime wait_abs;`, just an observation.

.github/workflows/ci.yml
```diff
@@ -182,4 +182,91 @@ jobs:
         with:
           name: MacOS CMake Test Logs
           path: build-aux/artifacts
-
```
JiaT75, line 185: The newline was deleted at the end of this file. It's just a minor thing, but can you please bring it back?

.github/workflows/ci.yml
```diff
@@ -182,4 +182,91 @@ jobs:
...
+        run: ./build-aux/ci_build.sh -b autotools -d decoders,shared -p test -n no_decoders
+
+      - name: Build without threads
+        shell: msys2 {0}
+        run: ./build-aux/ci_build.sh -b autotools -d threads,shared -p build
+      - name: Test without threads
+        shell: msys2 {0}
+        run: ./build-aux/ci_build.sh -b autotools -d threads,shared -p test -n no_threads
+
+      - name: Build without BCJ filters
```
JiaT75, line 239: Since the Windows CI takes so long, we can remove some of these less interesting tests. Let's omit the BCJ, Delta, and SHA256 trials. The CRC64 one could still be interesting, so lets keep that one.

JiaT75 (commented):
Overall, this is a great start and we appreciate your efforts! 

Eventually, it would be great to have CMake and some of the other msys2 flavors, but since the Windows tests are so slow they may belong in another workflow that does not run as often (Maybe just as a pre-release sanity check). Anyway, that is out of the scope of this PR.

arixmkii:
Thank you for detailed feedback! 👍  Will rework the changes and group them into logically separated commits.

src/liblzma/common/memcmplen.h
arixmkii: > Its interesting that this was only highlighted when threads are disabled.

This was really interesting. I tried to narrow it down, but didn't have juice to finish this investigation. Will apply recommended changes in a separate commit.

PR #42, opened 2023-02-26
project requires c99

autoantwort:
## Pull request checklist

## Pull request type

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

If the compiler uses a very old c standard as default, it will fail to build

Related Issue URL: https://github.com/microsoft/vcpkg/issues/21680#issuecomment-983342257

## What is the new behavior?
Set the required c standard to C99

- [ ] Yes
- [x] No

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -397,6 +397,7 @@ target_include_directories(liblzma PRIVATE
 )
 
 target_link_libraries(liblzma Threads::Threads)
+target_compile_features(liblzma PRIVATE c_std_99)
 
 # Put the tuklib functions under the lzma_ namespace.
 target_compile_definitions(liblzma PRIVATE TUKLIB_SYMBOL_PREFIX=lzma_)


JiaT75:
Hi! Thanks for the bug report and PR.

All of the targets that need to be compiled require C99, so it would be better to just set the CMAKE_C_STANDARD variable at the top of CMakeLists.txt like this:

```
set(CMAKE_C_STANDARD 99)
```

autoantwort:
Do the headers also require C99 or only the source code?

JiaT75:
The liblzma API headers should be compatible with C89, but the internal headers used by the source code are not.

JiaT75:
This should be fixed as of commit 4b7fb3bf41a0ca4c97fad3799949a2aa61b13b99 on master. @autoantwort can you let us know if this does not solve the bug?

This will be in a new stable 5.4.2 release in the near future. Thanks again for reporting this!

PR #51, opened 2023-05-31
protect against double find_package

bebuch:
Boost iostream [uses `find_package` in quiet mode](https://github.com/boostorg/iostreams/blob/5fe4de84f863964f7573be1146f524886146a5d3/CMakeLists.txt#L16) and then [again uses `find_package` with required](https://github.com/boostorg/iostreams/blob/5fe4de84f863964f7573be1146f524886146a5d3/CMakeLists.txt#LL30C1-L30C49). This second call triggers a `add_library cannot create imported target "ZLIB::ZLIB" because another target with the same name already exists.`

This can simply be fixed by skipping the alias part on secondary `find_package` runs.

## Pull request checklist

## Pull request type

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

From Boost iostream compilation which finds XZ via config file:

```
CMake Error at C:/Temp/lib/win_x86_64_debug/lib/cmake/zlib/ZLIBConfig.cmake:13 (add_library):
  add_library cannot create imported target "ZLIB::ZLIB" because another
  target with the same name already exists.
Call Stack (most recent call first):
  libs/iostreams/CMakeLists.txt:30 (find_package)
  libs/iostreams/CMakeLists.txt:38 (boost_iostreams_option)
```

## What is the new behavior?

`find_package` works on the first and all folloring runs.

- [ ] Yes
- [x] No

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -889,13 +889,15 @@ write_basic_package_version_file(
 set(LZMA_CONFIG_CONTENTS
 "include(\"\${CMAKE_CURRENT_LIST_DIR}/liblzma-targets.cmake\")
 
-# Be compatible with the spelling used by the FindLibLZMA module. This
-# doesn't use ALIAS because it would make CMake resolve LibLZMA::LibLZMA
-# to liblzma::liblzma instead of keeping the original spelling. Keeping
-# the original spelling is important for good FindLibLZMA compatibility.
-add_library(LibLZMA::LibLZMA INTERFACE IMPORTED)
-set_target_properties(LibLZMA::LibLZMA PROPERTIES
-                      INTERFACE_LINK_LIBRARIES liblzma::liblzma)
+if(NOT TARGET LibLZMA::LibLZMA)
+       # Be compatible with the spelling used by the FindLibLZMA module. This
+       # doesn't use ALIAS because it would make CMake resolve LibLZMA::LibLZMA
+       # to liblzma::liblzma instead of keeping the original spelling. Keeping
+       # the original spelling is important for good FindLibLZMA compatibility.
+       add_library(LibLZMA::LibLZMA INTERFACE IMPORTED)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -889,13 +889,15 @@ write_basic_package_version_file(
...
 
-# Be compatible with the spelling used by the FindLibLZMA module. This
-# doesn't use ALIAS because it would make CMake resolve LibLZMA::LibLZMA
-# to liblzma::liblzma instead of keeping the original spelling. Keeping
-# the original spelling is important for good FindLibLZMA compatibility.
-add_library(LibLZMA::LibLZMA INTERFACE IMPORTED)
-set_target_properties(LibLZMA::LibLZMA PROPERTIES
-                      INTERFACE_LINK_LIBRARIES liblzma::liblzma)
+if(NOT TARGET LibLZMA::LibLZMA)
+       # Be compatible with the spelling used by the FindLibLZMA module. This
```
JiaT75, line 893: There should be 4 spaces here to match the rest of the file

JiaT75:
Thanks the for PR! This is a great addition. It seems like the FindLibLZMA module has something similar to protect against multiple find_package() calls, so it seems like a common protection to include.

Can you wrap your commit message with newlines around 73 character width? The first sentence in your commit message is a bit long. The reason we care is because in our releases the Changelog is generated from the commit messages. In order to keep this and the commit log itself looking nice, we have a few standards we try to enforce.

bebuch:
@JiaT75 Unfortunately I don't know if this can be done online on GitHub and I'm too stressed right now to make an offline change using Git.

I have given you access to my fork. Would be nice if you could do that.

JiaT75:
@bebuch I was able to make the commit message change during the Squash and Merge. The commit is in your name and I only made stylistic edits to the message (I did not change the meaning).

Thanks for your contribution!

PR #52, opened 2023-06-02
tests/Makefile.am: add target of 'build-PROGRAMS'

duerpei:
Add target 'build-PROGRAMS' to only compile the test set under '/tests' and not run locally.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -76,6 +76,9 @@ if COND_SCRIPTS
 TESTS += test_scripts.sh
 endif
 
+build-PROGRAMS: all-am
+	$(MAKE) $(AM_MAKEFLAGS) $(check_PROGRAMS)
+
 clean-local:
 	-rm -f compress_generated_* \
 		xzgrep_test_output xzgrep_test_1.xz xzgrep_test_2.xz


JiaT75:
Hi! Thank you for the code suggestion. In order for us to accept a change like this, we need more information about the problem that is solves. What is the need to compile the tests programs locally without running them? Is it a cross-compile situation where you want to copy over the test binaries after everything else is built?

If this is the case, then a better solution is to override the TESTS variable in a make check command to be empty on your build machine:

`make check TESTS=`

The TESTS variable, from the Automake docs: 

"If the special variable TESTS is defined, its value is taken to be a list of programs or scripts to run in order to do the testing."

So, this is the list of tests to execute. If you leave it empty, it will still build all of the tests and then not execute any. Would this solve the issue?

duerpei:
> 

As you mentioned, it is a cross-compile situation where I want to copy over the test binaries after everything else is built. 
To be precise, I want to add "ptest" support for xz in "openembedded core", so there is the situations where only local compilation is done without running tests.

I have tried the method you suggested and it can also solve my problem very well, so there is no need to modify the code in xz.
Finally, I would like to thank you for your patient comment. 
Thank you!

duerpei:
My problem has been resolved, so I will close this issue.
Thank you for your reply.

PR #53, opened 2023-06-23
Replaced crc64_fast constructor with ifunc

hansjans162:
The ifunc attribute is a cleaner and marginally more performant
mechanism for dynamically choosing a routine at runtime.

The PR contains two parts:
1. The first replaces the crc64 construct with the ifunc resolver.
2. The second modifies the build process to account for the ifunc attribute.

HAVE_FUNC_ATTRIBUTE_IFUNC is set if ifuncs can be built
on the system. glibc has supported ifunc since version 2.11 (2009). Building for older machines
may want to manually disable ifunc support when building.
--disable-ifunc can be used with autotools and
USE_ATTR_IFUNC=OFF can be used with CMake to disable the check for ifunc support.
USE_ATTR_IFUNC defaults to ON for CMake and enable_ifunc defaults to yes for autotools.

More about ifuncs can be read [here](https://sourceware.org/glibc/wiki/GNU_IFUNC) or [here](https://www.agner.org/optimize/optimizing_cpp.pdf#page=142).

## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, no api changes)
- [X] Build related changes

crc64\_fast currently uses the \_\_attribute\_\_ constructor to choose the crc implementation to use.

## What is the new behavior?
I replaced the \_\_attribute\_\_((\_\_constructor\_\_)) with \_\_attribute\_\_((ifunc())) in crc64_fast.c

- The HAVE_FUNC_ATTRIBUTE_IFUNC flag was added to configure.ac
- The USE_ATTR_IFUNC cmake option was added to CMakeLists.txt

- [ ] Yes
- [X] No

Functionality should be identical.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -763,6 +763,25 @@ check_c_source_compiles("
 cmake_pop_check_state()
 tuklib_add_definition_if(liblzma HAVE_FUNC_ATTRIBUTE_CONSTRUCTOR)
 
+# Check for __attribute__((__ifunc__())) support.
+option(CHECK_ATTR_IFUNC "Use __attribute__((__ifunc__())) if supported by \
+the system" ON)
+
+if (CHECK_ATTR_IFUNC)
+    cmake_push_check_state()
+    set(CMAKE_REQUIRED_FLAGS "-Werror")
+    check_c_source_compiles("
+            static void func(void) { return; }
+            static void (*resolve_func (void)) (void) { return func; }
+            void func_ifunc (void)
+                    __attribute__ ((__ifunc__ (\"resolve_func\")));
+            int main(void) { return 0; }
+        "
+        HAVE_FUNC_ATTRIBUTE_IFUNC)
+    cmake_pop_check_state()
+    tuklib_add_definition_if(liblzma HAVE_FUNC_ATTRIBUTE_IFUNC)
+endif()
+
 # cpuid.h
 check_include_file(cpuid.h HAVE_CPUID_H)
 tuklib_add_definition_if(liblzma HAVE_CPUID_H)
--- a/src/liblzma/check/crc64_fast.c
+++ b/src/liblzma/check/crc64_fast.c
@@ -438,22 +438,25 @@ is_clmul_supported(void)
 }
 
 
-#ifdef HAVE_FUNC_ATTRIBUTE_CONSTRUCTOR
-#	define CRC64_FUNC_INIT
-#	define CRC64_SET_FUNC_ATTR __attribute__((__constructor__))
-#else
-#	define CRC64_FUNC_INIT = &crc64_dispatch
-#	define CRC64_SET_FUNC_ATTR
-static uint64_t crc64_dispatch(const uint8_t *buf, size_t size, uint64_t crc);
-#endif
+#if defined(HAVE_FUNC_ATTRIBUTE_IFUNC)
+// The resolver must accept no arguments (void), and must specify the return
+// and parameter types in order to avoid warnings. The resolved function return
+// type appears where a regular function's return type would be, and the
+// resolved functions parameters appear after the resolver's arguments (void).
+static uint64_t
+(*crc64_resolver(void)) (const uint8_t*, size_t, uint64_t)
+{
+	return is_clmul_supported() ? crc64_clmul : crc64_generic;
+}
 
+// The resolved function can be called as crc64_func().
+static uint64_t crc64_func(const uint8_t *buf, size_t size, uint64_t crc)
+		__attribute__ ((__ifunc__ ("crc64_resolver")));
 
-// Pointer to the the selected CRC64 method.
-static uint64_t (*crc64_func)(const uint8_t *buf, size_t size, uint64_t crc)
-		CRC64_FUNC_INIT;
+#elif defined(HAVE_FUNC_ATTRIBUTE_CONSTRUCTOR)
+static uint64_t (*crc64_func)(const uint8_t *buf, size_t size, uint64_t crc);
 
-
-CRC64_SET_FUNC_ATTR
+__attribute__((__constructor__))
 static void
 crc64_set_func(void)
 {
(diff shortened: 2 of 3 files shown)


JiaT75:
Thanks for the PR and the helpful links! Overall this seems like a nice improvement to our function picking strategy for CRC64. It will likely be useful when we implement CRC32 CLMUL too :)

I want to do a bit more research about the GNU indirect functions to make its right for this project before we decide to merge. Before that, there are a few style changes that need to be made. I will comment them separately.

CMakeLists.txt
```diff
@@ -763,6 +763,31 @@ check_c_source_compiles("
 cmake_pop_check_state()
 tuklib_add_definition_if(liblzma HAVE_FUNC_ATTRIBUTE_CONSTRUCTOR)
 
+# Check for __attribute__((ifunc())) support.
+option(USE_ATTR_IFUNC "Use __attribute__((ifunc())) if supported by the system"
+        ON)
+
+if (USE_ATTR_IFUNC)
```
JiaT75, line 770: This option name is a bit misleading. Enabling it will only check for __attribute__((ifunc())) support, not guarantee they are used. Maybe rename it to CHECK_ATTR_IFUNC?

src/liblzma/check/crc64_fast.c
```diff
@@ -438,35 +438,47 @@ is_clmul_supported(void)
...
-#endif
+#ifdef HAVE_FUNC_ATTRIBUTE_IFUNC
+#	ifdef __cplusplus
+// The extern variable "C" is required to work with C++'s function name 
+// mangling in C++. If this is not included, the "crc64_resolver" will be
+// mangled and won't compile.
+extern "C" {
+#	endif
 
+// The resolver must accept no arguments (void), and must specify the return and
```
JiaT75, line 449: We try to keep an 80 character width in this project. Can you wrap these comment lines at 80?

src/liblzma/check/crc64_fast.c
```diff
@@ -438,35 +438,47 @@ is_clmul_supported(void)
...
-#	define CRC64_FUNC_INIT
-#	define CRC64_SET_FUNC_ATTR __attribute__((__constructor__))
-#else
-#	define CRC64_FUNC_INIT = &crc64_dispatch
-#	define CRC64_SET_FUNC_ATTR
-static uint64_t crc64_dispatch(const uint8_t *buf, size_t size, uint64_t crc);
-#endif
+#ifdef HAVE_FUNC_ATTRIBUTE_IFUNC
+#	ifdef __cplusplus
+// The extern variable "C" is required to work with C++'s function name
```
JiaT75, line 443: There is a trailing space at the end of this line

Larhzu:
My understanding is that ifunc is specific to glibc. `__constructor__` works on many ELF platforms, not just GNU/Linux.

According to your link, ifunc is incompatible with `-fsanitize=address`. And indeed it makes `make check` fail with segfaults. (I also see that two tests fail with `-fsanitize=address` before your patch due to memory leaks but no segfaults.)

Also, checking for `__cplusplus` isn't needed as this is C code and not a public header.

How much difference (speed or memory or anything else) does ifunc make in this specific use case? I understand that the function pointer is an extra step on each call but the function itself is more expensive than `memcpy` or such which get called a lot with tiny buffers. Currently I feel the extra complication isn't worth it in liblzma as `__constructor__` is good to keep around too for ELF platforms that lack ifunc support.


Thanks!

hansjans162:
Thanks for the feedback!

I could make the selection method a three tiered approach where ifunc is preferred, followed by the `__constructor__`, and followed again by the generic method. This way the behavior would remain unchanged on non-GNU/Linux systems. Yes, if `-fsantize=address` is set then you would want to set `--disable-ifunc`.

The performance improvements are helpful if crc64 is called a large amount, e.g. for a large number of small files or a multi-threaded file with many blocks. I'm using the crc64 function a very large amount by itself since I noticed it is very fast for checksumming since the CLMUL update. There is also some benefit with branch prediction and it gave me one less instruction when I compiled it with GCC. Your right though, I was hoping to look into additional optimizations after this one.

In regards to `__cplusplus` I wasn't sure if liblzma was being compiled with a C++ compiler anywhere. I can remove it if you don't think it's useful.

Larhzu:
I understand that ifunc avoids loading a function pointer and an indirect branch. liblzma uses function pointers a lot so, for (de)compression performance, avoiding one pointer in the call stack likely doesn't matter (but I haven't benchmarked it). Many `lzma_crc64` calls can process a few kilobytes of data per call but I understand that in non-compression uses tiny buffers may be the most common.

How big difference in speed does your patch make with your code? I would like understand the real-world improvement that ifunc can make specifically with `lzma_crc64`.

> In regards to __cplusplus I wasn't sure if liblzma was being compiled with a C++ compiler anywhere. I can remove it if you don't think it's useful.

C++ compiler isn't used so the check should be removed.

By the way, have you checked xxhash if you need a very fast hash?

Thanks!

hansjans162:
I have a large number of varying buffer sizes. With the larger buffers it doesn't make much of a difference, but with the smaller ~64B to ~256B buffers I was noticing a 4-5% improvement. I'm also running all of this on older hardware which may be contributing to the speedup. I benchmarked it for the standard .xz case, however the speedup there was within the bounds of being noise.

liblzma works well since it's already present, and we're also locked into the existing checksums.

Larhzu:
4-5 % isn't a huge difference but it's around the threshold where it becomes interesting. Jia and I discussed this and ifunc support is welcome while keeping the constructor attribute as the second choice. Thanks!

src/liblzma/check/crc64_fast.c
```diff
@@ -438,35 +438,47 @@ is_clmul_supported(void)
...
+
+#	ifdef __cplusplus
+}
+#	endif
 
-// Pointer to the the selected CRC64 method.
-static uint64_t (*crc64_func)(const uint8_t *buf, size_t size, uint64_t crc)
-		CRC64_FUNC_INIT;
+// The resolved function can be called as crc64_func().
+static uint64_t crc64_func(const uint8_t* buf, size_t size, uint64_t crc)
```
JiaT75, line 464: To keep the style consistent, please change the pointer declarations to be type *var_name instead of type* var_name

hansjans162:
I made all of the requested fixes. Let me know if there are any other concerns.

Larhzu:
When fixing commits in a patchset / pull request, it usually should be done by editing the original commits instead of adding fix-up commits at the end. For example, in this case the commits first remove a feature and then add it back. This makes it harder to see what changes were _actually_ made in the end. This applies to both reviewing the changes before the merge and to people who read xz's commit log afterwards. I think this should be fixed.

Thanks!

hansjans162:
Thanks for the advice. I made the requested change. I rebased the git commits and updated the commit messages.

Larhzu:
Using ifunc for a static `crc64_func` means that `lzma_crc64` becomes a single-instruction function that just does a jump via PLT:

```
jmp    45a0 <*ABS*+0x15e00@plt>
```

I suppose it's better to make `lzma_crc64` itself the ifunc. This has been done in [v2](https://github.com/tukaani-project/xz/tree/ifunc-crc64-fast-v2) branch. Can you test it and tell if you notice any difference (speed or anything else) compared to your version. Thanks!

JiaT75:
@hansjans162 Just merged from the other branch Thanks for your contribution!

RashadGasimli:
wtf

PR #63, opened 2023-09-15
Fix getopt.in.h for MSVC

kiyolee:
## Pull request checklist

## Pull request type

 

- [ ] Bugfix
- [ ] Feature
- [X] Build related changes

cmake build with MSVC failed as "getopt.in.h" included <unistd.h> which does not exist for MSVC.

Related Issue URL: 

## What is the new behavior?

cmake build with MSVC succeeded.

- [ ] Yes
- [X] No

### DIFF
--- a/lib/getopt.in.h
+++ b/lib/getopt.in.h
@@ -34,7 +34,9 @@
 #if defined __GETOPT_PREFIX && !defined __need_getopt
 # include <stdlib.h>
 # include <stdio.h>
+#ifndef _MSC_VER
 # include <unistd.h>
+#endif
 # undef __need_getopt
 # undef getopt
 # undef getopt_long


kiyolee:
Would you consider renaming "getopt.in.h" to "getopt.h"?
There is nothing in "getopt.in.h" that needs to be generated.
The generated "getopt.h" by cmake build is exactly the same as "getopt.in.h".

Larhzu:
> Would you consider renaming "getopt.in.h" to "getopt.h"?

getopt.in.h is done like that in Gnulib. A package that uses Gnulib may have many .in.h files. Which files will be needed is detected when running `configure` and copied to .h name. Gnulib has replacements for many system headers to aid portability. With the .in.h -> .h method only the specificic headers can be overriden.

Obviously XZ Utils only include getopt from Gnulib at the moment. In the early days I didn't know if more modules would be needed. So moving the getopt files to lib/getopt/ and putting that to include path when needed could be fine if we are certain that the module list won't grow. If many Gnulib modules (or similar things from other sources) were needed then this wouldn't work because the include path would grow too long and the modules can have intermodule dependencies too.

I will discuss this with Jia. We plan to update getopt code with the current Gnulib too (it's still LGPLv2.1 so no license changes).

I have seen your other messages. I will get back to them later this week.

Larhzu:
The getopt.in.h change is in xz_for_msvc branch now. Thanks!

PR #73, opened 2023-11-28
Improve existing oss-fuzz coverage

mvatsyk-lsg:
## Pull request checklist

- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [ ] Build was run locally and without warnings or errors

## Pull request type

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Related Issue URL: N/A

## What is the new behavior?

- Total number of fuzzers increased from 1 to 12
- 6 new fuzzers for decompressing supported archive types
- 5 new fuzzers for compressing data into supported archive types
- Added LZMA Raw test files based on the existing `.lzma` ones

- [x] Yes
- [ ] No

This pull request will temporarily break existing oss-fuzz setup until the `oss-fuzz` repo accepts [a corresponding pull request](https://github.com/google/oss-fuzz/pull/11279) with the updated fuzzer configuration on their side.

## Other information

The improvements to the fuzzing setup were made as a part of Google ISE project.


JiaT75:
Hello!

Thanks for the PR. This is a great start to improving the fuzz testing. I will start with a few overall comments here and then add some more specific comments directly on the commits themselves.

First, we need to be sure that we are using the fuzz resources in the best way we can. Its easy to think of the OSS-Fuzz resources as unlimited, but each project can only be fuzzed so much. We should only include a fuzz target if it provides clear value and is testing an important part of liblzma that isn't being covered by a different fuzz target. Otherwise, less useful fuzz targets will take away compute time from the more useful ones. So, can you justify the reasoning behind each of the new fuzz targets? For instance, I am not sure that the raw encoder and decoder fuzz targets are useful since their important code paths are already covered by every other fuzz target. The raw coders don't have important header data, its just raw LZMA data. I am likely missing an important fuzz case, but in my mind I can think of three useful things to fuzz in our library:

- Metadata encoding/decoding (magic bytes, file headers, block headers, lzma2 chunk headers, etc.)
- Filter data encoding/decoding (LZMA1, BCJ, delta)
- Check functions (CRC32, CRC64, SHA256)

Next, the code itself has a lot of repeated boilerplate. Each of the fuzz targets has very little unique code. For instance, this could be reorganized into a shared header file that provides a function for encoding and a function for decoding. These functions can take the coder init function (lzma_alone_decoder(), lzma_auto_decoder(), etc.) as a function pointer arg and any needed flags or options.

We could also consider fuzzing the various BCJ filters (x86, PowerPC, ARM64, etc). These filters are designed to be applied to executable data, but will be run on non-executable data very often. So its possible that there are hidden data corruption bugs on an unexpected input sequence since they are mostly tested on executable data, making it a good candidate for fuzz testing. These filters cannot be used as raw coders at this time, so they will have to be combined in a filter chain with LZMA1/2. If we want to look for data corruption bugs, we should encode a chunk, then decode it and compare if the decoded version exactly matches the original data.

For your commit messages, we like to keep a consistent format. When we release, our Changelog is generated automatically from the contents of the commit messages. Also it helps us maintain our codebase better when the commit messages are descriptive and clear. For your commits, please have them start with the category of what they are changing. For these, I would prepend "Tests:" to the first line of each commit. The first line of each commit should be a brief description of the purpose of the commit. The following lines should explain what was changed and why. Make sure to wrap the lines of the commit message to at most 73 characters since different commit log viewers may or may not wrap long lines and it helps keep a consistent look in our Changelog.

tests/ossfuzz/config/fuzz_lzip.dict
```diff
@@ -0,0 +1 @@
+"LZIP"
```
JiaT75, line 1: This dictionary could also include valid lzip header bytes for versions 0 and 1. See the [file format](https://www.nongnu.org/lzip/lzip_talk_ghm_2019.html) for what these bytes should be.

tests/ossfuzz/config/fuzz_decode_raw_lzma2.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
+dict = fuzz_xz.dict
```
JiaT75, line 2: Raw LZMA2 does not use the .xz magic bytes, so this dictionary isn't that helpful for the fuzz test

tests/ossfuzz/fuzz_decode_stream.c
```diff
@@ -1,10 +1,13 @@
 ///////////////////////////////////////////////////////////////////////////////
 //
-/// \file       fuzz.c
+/// \file       fuzz_decode_stream.c
 /// \brief      Fuzz test program for liblzma
 //
 //  Author:     Lasse Collin
 //
+//  Original fuzz.c fuzzer, renamed to fuzz_decode_stream to conform with new
```
JiaT75, line 8: There isn't a need to mention the previous filename. This information is capture in the commit history

tests/ossfuzz/Makefile
```diff
@@ -1,6 +1,61 @@
-fuzz: fuzz.c
-	$(CC) $(CFLAGS) -c fuzz.c -I ../../src/liblzma/api/
-	$(CXX) $(CXXFLAGS) $(LIB_FUZZING_ENGINE) fuzz.o -o $(OUT)/fuzz \
+fuzz_decode_xz: fuzz_decode_xz.c
+	$(CC) $(CFLAGS) -c fuzz_decode_xz.c -I ../../src/liblzma/api/
+	$(CXX) $(CXXFLAGS) $(LIB_FUZZING_ENGINE) fuzz_decode_xz.o -o $(OUT)/fuzz_decode_xz \
+	      ../../src/liblzma/.libs/liblzma.a
+
+fuzz_decode_xz_crc: fuzz_decode_xz_crc.c
```
JiaT75: This Makefile could be done with wildcards instead of repeating the same commands for each fuzz target since all of the .c/.o files become fuzz targets.

mvatsyk-lsg:
@JiaT75 thanks a lot for the swift review! I will start implementing the suggested changes asap. I am going to focus on fixing the highlighted issues before implementing any new fuzzers, if that makes sense :) 

To address some of your comments above:
- The new fuzz targets were chosen to cover most of the `liblzma` common use cases from the code of XZ utils (i.e. we expanded the tests to cover the decoders and encoders used by the CLI utilities) 
- I completely agree that the fuzz targets' code can be reduced via templates and shared code. Would you like to simply move the common functionality into a separate header file, or to generate the fuzzers' code dynamically by replacing the lines in a template file as well?

JiaT75:
> @JiaT75 thanks a lot for the swift review! I will start implementing the suggested changes asap. I am going to focus on fixing the highlighted issues before implementing any new fuzzers, if that makes sense :)

Makes perfect sense. I noted the BCJ filter fuzzing as option to consider. We don't necessarily need to implement it or implement it right away. Just an idea of something we could also be fuzzing if we agree the value is there.

> * I completely agree that the fuzz targets' code can be reduced via templates and shared code. Would you like to simply move the common functionality into a separate header file, or to generate the fuzzers' code dynamically by replacing the lines in a template file as well?

I think the simplest approach would be to use a common separate header file. Creating a dynamic template would take some extra build logic whereas an extra header file would only require updating the Makefile.

mvatsyk-lsg:
@JiaT75 I've addressed your comments and tested the new Makefile and fuzz targets' code on a local setup. They seem to compile and work just fine. A quick question: should I go back and edit the description of all commits before the review, or will you be making a squash during the merge?

JiaT75:
> @JiaT75 I've addressed your comments and tested the new Makefile and fuzz targets' code on a local setup. They seem to compile and work just fine. A quick question: should I go back and edit the description of all commits before the review, or will you be making a squash during the merge?

We like to keep our commits small and focused, so we will likely want more than one commit for this many changes. For now, don't worry about squashing your commits until the review is basically done. At the end we can figure out how many commits are appropriate for this and squash accordingly. So feel free to keep adding fix up commits as we go.

I'll start reviewing your new changes.

tests/ossfuzz/config/fuzz_decode_auto.options
JiaT75, line 1: Does every fuzz target need a `.options` file? If not, then I don't think we need to include a `.options` file if the fuzz target doesn't need anything special

tests/ossfuzz/config/fuzz_lzma.dict
```diff
@@ -0,0 +1,20 @@
+# first 5 header bytes of .lzma archives based on the info from
+# - extra/scanlzma/scanlzma.c (lines 32-52)
+# - src/liblzma/common/alone_encoder.c (lines 107-126)
+# - https://github.com/cscott/lzma-purejs/blob/master/FORMAT.md
+# - manual hexdump of data produced by the lzma CLI suite
+
+# compression mode 1    (LZMA properties = 0x5d, dict size = 0x00100000)
```
JiaT75, line 7: Instead of **mode** it should be **preset**. In the documentation **mode** in the context of compression mode refers to a match finder setting (fast or normal).

The LZMA properties resulting in 0x5d are lc=3, lp=0, pb=2. It could be helpful to mention this once, and then only mention the dictionary size for each .dict entry since the LZMA properties aren't changing.

tests/ossfuzz/fuzz_common.h
```diff
@@ -0,0 +1,43 @@
+#include <inttypes.h>
```
JiaT75, line 1: The header comments are missing from this file

tests/ossfuzz/fuzz_common.h
```diff
@@ -0,0 +1,43 @@
...
+
+
+// Output buffer for decompressed data. This is write only; nothing cares
+// about the actual data written here.
+static uint8_t OUTBUF[4096];
+
+// Some header values can make liblzma allocate a lot of RAM
+// (up to about 4 GiB with liblzma 5.2.x). We set a limit here to
+// prevent extreme allocations when fuzzing.
+const uint64_t MEM_LIMIT = 300 << 20; // 300 MiB
```
JiaT75, line 14: This variable should also be static (`static const ...`)

tests/ossfuzz/fuzz_common.h
```diff
@@ -0,0 +1,43 @@
...
+
+
+void fuzz_code(lzma_stream *stream, const uint8_t *inbuf, size_t inbuf_size) {
```
JiaT75, line 17: The function name should start on a new line. And this should be static. So the first line:
static void

Second line:
fuzz_code(...

This is to match our style in the project. Having the function name starting at column index 0 helps certain tools and text editors search for function names.

tests/ossfuzz/fuzz_common.h
```diff
@@ -0,0 +1,43 @@
...
+
+// Some header values can make liblzma allocate a lot of RAM
+// (up to about 4 GiB with liblzma 5.2.x). We set a limit here to
+// prevent extreme allocations when fuzzing.
+const uint64_t MEM_LIMIT = 300 << 20; // 300 MiB
+
+
+void fuzz_code(lzma_stream *stream, const uint8_t *inbuf, size_t inbuf_size) {  
+    // Give the whole input buffer at once to liblzma.
+	// Output buffer isn't initialized as liblzma only writes to it.
```
JiaT75, line 19: It looks like there is a mix of tabs and spaces. In .c files we use tabs, and treat the tab width as 8 characters. I'm sure other projects do this differently since there is no right answer to the tabs vs spaces debate

tests/ossfuzz/fuzz_common.h
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	// LZMA_PROG_ERROR should never happen as long as the code calling
+	// the liblzma functions is correct. Thus LZMA_PROG_ERROR is a sign
+	// of a bug in either this function or in liblzma.
+	if (ret == LZMA_PROG_ERROR) {
+		fprintf(stderr, "lzma_code() returned LZMA_PROG_ERROR\n");
+		abort();
+	}
+}
```
JiaT75, line 43: Missing an empty newline at the end of the file

tests/ossfuzz/fuzz_common.h
```diff
@@ -0,0 +1,43 @@
...
+    // Give the whole input buffer at once to liblzma.
+	// Output buffer isn't initialized as liblzma only writes to it.
+	stream->next_in = inbuf;
+	stream->avail_in = inbuf_size;
+	stream->next_out = OUTBUF;
+	stream->avail_out = sizeof(OUTBUF);
+
+    lzma_ret ret;
+	while ((ret = lzma_code(stream, LZMA_FINISH)) == LZMA_OK) {
+		if (stream->avail_out == 0 || ret == LZMA_STREAM_END) {
```
JiaT75, line 27: There is no need to check `ret == LZMA_STREAM_END` here. The while loop would end if `ret` isn't `LZMA_OK` here.

tests/ossfuzz/fuzz_decode_alone.c
```diff
@@ -0,0 +1,41 @@
...
+#include "fuzz_common.h"
+
+
+extern int
+LLVMFuzzerTestOneInput(const uint8_t *inbuf, size_t inbuf_size)
+{
+	// Initialize a LZMA alone decoder using the above memory usage limit.
+	lzma_stream strm = LZMA_STREAM_INIT;
+	lzma_ret ret = lzma_alone_decoder(&strm, MEM_LIMIT);
+	if (ret != LZMA_OK) {
```
JiaT75, line 28: Since `ret` isn't used outside of this if statement anymore, its enough to just say:

if (lzma_alone_decoder(&strm, MEM_LIMIT) != LZMA_OK)

and remove the `ret` variable from these programs.

tests/ossfuzz/fuzz_decode_auto.c
```diff
@@ -0,0 +1,52 @@
...
+#include <stdlib.h>
+#include <stdio.h>
+#include "lzma.h"
+#include "fuzz_common.h"
+
+
+extern int
+LLVMFuzzerTestOneInput(const uint8_t *inbuf, size_t inbuf_size)
+{
+	// Initialize an Auto decoder using the above memory usage limit.
```
JiaT75, line 25: The memory limit has moved to the header file so this comment should be updated

tests/ossfuzz/fuzz_decode_raw.c
JiaT75, line 1: I'm not convinced this fuzz target is needed. The interesting LZMA decoding code should be covered by `fuzz_decode_alone` already.

tests/ossfuzz/fuzz_decode_raw_lzma2.c
```diff
@@ -0,0 +1,52 @@
...
+
+	// Initialize a raw decoder using the above filter chain.
+	lzma_stream strm = LZMA_STREAM_INIT;
+	lzma_ret ret = lzma_raw_decoder(&strm, filters);
+	
+	if (ret != LZMA_OK) {
+		// This should never happen unless the system has
+		// no free memory or address space to allow the small
+		// allocations that the initialization requires.
+		fprintf(stderr, "lzma_raw_decoder() LZMA 2 failed (%d)\n", ret);
```
JiaT75, line 42: Since we treat tabs at 8 character width, this line is more than 80 characters. We try to always wrap our lines at 80 characters

tests/ossfuzz/fuzz_encode_raw_lzma2.c
```diff
@@ -0,0 +1,53 @@
+///////////////////////////////////////////////////////////////////////////////
```
JiaT75, line 1: We can probably omit this fuzz target since its interesting code paths will be hit by `fuzz_encode_stream`

tests/ossfuzz/fuzz_encode_raw.c
```diff
@@ -0,0 +1,53 @@
+///////////////////////////////////////////////////////////////////////////////
```
JiaT75, line 1: We can also omit this fuzz target. It should be covered by `fuzz_encode_alone`, which is mostly covered by `lzma_encode_stream()` anyway

tests/ossfuzz/config/fuzz_decode_alone.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
+dict = fuzz_lzma.dict
```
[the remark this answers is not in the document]
mvatsyk-lsg: I believe the oss-fuzz will fall back to the default max length of the generated input of 4096 bytes if the `max_len` option is not provided in the `.options` file, so setting it explicitly is not necessary. My bad for not standardising the options file formats. 

We can either remove the files entirely as per the above comment, or include this option explicitly in each config file. Which options sounds better to you?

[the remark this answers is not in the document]
JiaT75: Based on the [libFuzzer docs](https://llvm.org/docs/LibFuzzer.html#options), it seems that the default is a special 0 value that indicates the maximum input length should be inferred from the corpus. Most of our test files are small (I believe the largest is ~50 Kb .xz file) so its likely that the default value is good enough in practice.

However it could be a good idea to set an upper bound on a size we think is reasonable, especially for the fuzz targets that are meant to test handling bytes early in the input stream. So I would vote for this.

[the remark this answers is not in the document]
mvatsyk-lsg: Good! What size would you consider reasonable for the test cases?

mvatsyk-lsg:
A follow up on the redundant fuzzers: I ran the setup without them, and the coverage difference is indeed negligible. I am removing them from the pull request

tests/ossfuzz/fuzz_decode_alone.c
JiaT75: I apologize, this was a bad suggestion from me. I didn't realize the `ret` value was being used in the error message, which could be helpful if the fuzz targets somehow fail when creating the coders. Don't worry about reverting this, I will make the change after we merge.

mvatsyk-lsg:
Apart from the possible re-addition of `options` files with `max_len` options back to the pull request, everything is ready for your review @JiaT75 !

mvatsyk-lsg:
Also, I am now questioning whether the addition of `.lzma_raw` files is needed, since the corresponding fuzzers were removed from the pull request

tests/ossfuzz/config/fuzz_decode_alone.options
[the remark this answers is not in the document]
JiaT75: > Good! What size would you consider reasonable for the test cases?

The 4096 number you were using before seems good. That should be more than enough to trigger the sanitizers if there is a possible bad code path

mvatsyk-lsg:
> > Also, I am now questioning whether the addition of `.lzma_raw` files is needed, since the corresponding fuzzers were removed from the pull request
> 
> Its safe to remove the `.lzma_raw` files and the `tests/files/README` changes.
> 
> Thanks for all the changes so far! I feel we are getting close to this being ready.

Thank you for the review! I've reverted the changes and added the `max_len=4096` to all fuzzer options.

tests/ossfuzz/fuzz_decode_alone.c
```diff
@@ -0,0 +1,41 @@
+///////////////////////////////////////////////////////////////////////////////
+//
+/// \file       fuzz_decode_alone.c
+/// \brief      Fuzz test program for liblzma lzma_alone_decoder()
+//
+//  Author:     Maksym Vatsyk
+//
+//  Based on Lasse Collin's original fuzz.c fuzzer for liblzma
```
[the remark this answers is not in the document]
Larhzu: Just listing the names in "Authors:" works best. I suspect that I didn't write the exact fuzz.c but it's based on the example programs in doc/examples which I did write. With these particular files, the author info doesn't matter to me, it's just that someone auditing where the code came from might care about "Authors:" convering everyone.

Larhzu:
Thanks to both of you for your work so far!

There are a few things I would like to understand better. I have only skimmed OSS-Fuzz's docs so I might be asking silly questions, sorry.

1. Seems that [renaming a fuzz target](https://google.github.io/oss-fuzz/faq/#what-happens-when-i-rename-a-fuzz-target-) requires renaming the accumulated corpora too.

2. Does adding more fuzzers mean that the project-specific fuzzing resources (processor time) will be divided between the fuzzers? With a quick look I didn't find any advice about resource usage in OSS-Fuzz docs and it's not discussed much in this thread either.

3. The value of code coverage in fuzzing is unclear. *If* extending coverage by a few simple lines of code could slow down fuzzing of more important parts of the code, does it make sense to extend fuzzing coverage in that case? I'm thinking of cases where an old-school code review shouldn't take a lot of time (code snippets that are about 200 lines each and do nothing unusually complicated). Or perhaps these should be fuzzed at first but disabled after some time if they find nothing?

Examples of remaining significant overlap in the new fuzzing targets:

* fuzz_encode_alone.c would test end of payload marker (EOPM) encoding in LZMA but otherwise it doesn't test much that won't be tested by fuzz_encode_stream.c. They both use the LZMA encoder in the end. So it seems that fuzz_encode_alone.c isn't useful and could _maybe_ even be harmful due to resource usage unless the fuzzers are smart enough to spot when code paths become identical.

* fuzz_decode_alone.c splits into three different decoders depending on the input. Yet the three decoders are fuzzed separately too (stream, alone, lzip). So the only extra fuzzed thing is the small auto_decoder.c.

I don't know enough about the fuzzing methods to know what actually makes sense. I would like to be assured that adding all these fuzzers adds real value.

Thanks!

mvatsyk-lsg:
Hi @Larhzu !

> Seems that [renaming a fuzz target](https://google.github.io/oss-fuzz/faq/#what-happens-when-i-rename-a-fuzz-target-) requires renaming the accumulated corpora too.

In the existing setup, the corpora are generated dynamically in a [build.sh](https://github.com/google/oss-fuzz/pull/11279/commits/97ba2c05158912b2c8a5a2dd6c721fa31f2ed819) file. So, any modifications have to be done in a separate [pull request](https://github.com/google/oss-fuzz/pull/11279). After we merge this pull request, I will go ahead and update the latter one to properly reflect all the changes. 

> Does adding more fuzzers mean that the project-specific fuzzing resources (processor time) will be divided between the fuzzers? With a quick look I didn't find any advice about resource usage in OSS-Fuzz docs and it's not discussed much in this thread either.

I did not find any hard cap of the execution time for the OSS Fuzz itself. However, their CI integration, CIFuzz will divide the [shared fuzzing time of 10 minutes](https://google.github.io/oss-fuzz/getting-started/continuous-integration/#how-it-works) between all fuzz targets in the project. 

Getting back to the OSS Fuzz, each fuzz target will be run on [a dedicated machine](https://google.github.io/oss-fuzz/faq/#what-are-the-specs-on-your-machines) with 1 CPU and a cap of ~ 2GB RAM. 

Since the fuzzers are written in C/C++, I doubt that introducing new fuzzers, at least for now, will decrease the overall quality of the fuzzing output. On my test setup inside a VM with similar hardware parameters, the fuzzing and the generation of an introspector report took around 5 minutes.

> The value of code coverage in fuzzing is unclear. If extending coverage by a few simple lines of code could slow down fuzzing of more important parts of the code, does it make sense to extend fuzzing coverage in that case?

This absolutely makes sense. However, current fuzzing setup is very limited and covers only half of the lib (since `--disable-encoders` flag is used during the compilation). Its runtime coverage is 116/162 functions. The setup proposed in this pull request extends the fuzzing coverage to all common encoders and decoders to increase the runtime coverage to 270/360 functions.

> fuzz_encode_alone.c would test end of payload marker (EOPM) encoding in LZMA but otherwise it doesn't test much that won't be tested by fuzz_encode_stream.c. They both use the LZMA encoder in the end. So it seems that fuzz_encode_alone.c isn't useful and could maybe even be harmful due to resource usage unless the fuzzers are smart enough to spot when code paths become identical.

> fuzz_decode_alone.c splits into three different decoders depending on the input. Yet the three decoders are fuzzed separately too (stream, alone, lzip). So the only extra fuzzed thing is the small auto_decoder.c.

We can remove those, however this decreases the runtime fuzzing coverage from 270/360 to 249/360 functions. Should we proceed with deleting the fuzz targets?

mvatsyk-lsg:
Okay, @JiaT75, I've rebased the pull request. Does the commit history look good to you?

DavidKorczynski:
Thanks for reaching out @mvatsyk-lsg -- I didn't go through the whole discussion here so am trying to give an OSS-Fuzz perspective from a limited understanding of this PR.

Regarding OSS-Fuzz resources, then I think by default it makes sense to not be too concerned about this. OSS-Fuzz relies on [Clusterfuzz](https://github.com/google/clusterfuzz) which has a set of scheduling/prioritisation strategies. A single fuzzer for CRC may be a bit much. However, it's also possible to merge a bunch of simple fuzzers into a single larger function:

```cpp
int LLVMFuzzerTestOneInput(uint8_t *data, size_t size) {

  if (size < 1) {
    return 0;
  }
  uint8_t decider = data[0];
  data++;
  size--;
  switch decider {
      case 1: { fuzz_first_entrypoint(data, size); break; }
      case 2: { fuzz_second_entrypoint(data, size); break; }
      ...
      case N
    }
}
```

This is often a common strategy for hitting smaller functions. In fact, you can even do this by throwing the same smaller fuzzers into the larger meaningful fuzzers -- the fuzzer will through it's mutational genetic algorithm start exploring the code where there is more code to explore, so more efforts will be "put in the right places".

The scheduling in Clusterfuzz will be responsible for dividing time allocated to each of the targets.

JiaT75:
Thank you @DavidKorczynski for the explanation and the advice for combining fuzzers. I had not thought of using a byte from the fuzz input to control the fuzzer's entry point.

With that in mind, @mvatsyk-lsg we should combine `fuzz_encode_stream.c` and `fuzz_encode_stream_light.c` into just one fuzzer. We can use the same name `fuzz_encode_stream.c` for this fuzz target. We can use the first byte of input to help us determine the preset level. So the function could look like:

```c
extern int
LLVMFuzzerTestOneInput(const uint8_t *inbuf, size_t inbuf_size)
{
	if (size == 0)
		return 0;

	lzma_stream strm = LZMA_STREAM_INIT;

	uint32_t preset_level;

	uint8_t decider = inbuf[0];

	switch (decider) {
	case 0:
	case 1:
	case 5:
		preset_level = decider;
		break;
	case 6:
		preset_level = 0 | LZMA_PRESET_EXTREME;
		break;
	case 7:
		preset_level = 3 | LZMA_PRESET_EXTREME;
		break;
	}

	lzma_options_lzma opt_lzma;
	if (lzma_lzma_preset(&opt_lzma, preset_level)){
		fprintf(stderr, "lzma_lzma_preset() failed\n");
		abort();
	}

	// Initialize filter chain for lzma_stream_decoder() call
	// Use single LZMA2 filter for encoding
	lzma_filter filters[2];
	filters[0].id = LZMA_FILTER_LZMA2;
	filters[0].options = &opt_lzma;
	filters[1].id = LZMA_VLI_UNKNOWN;

	// Initialize the stream encoder using the above
	// filter chain and CRC64.
	if (lzma_stream_encoder(&strm,
			filters, LZMA_CHECK_CRC64) != LZMA_OK) {
		fprintf(stderr, "lzma_stream_encoder() failed\n");
		abort();
	}

	fuzz_code(&strm, inbuf, inbuf_size);

	// Free the allocated memory.
	lzma_end(&strm);
	return 0;
}
```

This can test a few different dictionary sizes, match finders, nice length, modes, and depth levels with the same fuzzer.

mvatsyk-lsg:
@JiaT75 done!

JiaT75:
@mvatsyk-lsg Thanks! Things are looking pretty good now. I created a separate [branch](https://github.com/tukaani-project/xz/tree/oss_fuzz) with all of your commits plus a minor cleanup commit. Can you test this branch to be sure I didn't break anything during my changes?

As I'm sure you know, the draft PR in OSS-Fuzz needs updating from all the changes we made here. I want to do a quick local test before merging but it will be easier if the OSS-Fuzz changes are updated on that PR.

mvatsyk-lsg:
@JiaT75 the changes look good to me! I have also updated the pull request to the oss-fuzz repo. To test the new fuzzing setup locally, you can run the following commands on your machine:

```bash
# clone my fork of the oss-fuzz repo with pull request changes
git clone https://github.com/mvatsyk-lsg/oss-fuzz
cd oss-fuzz/

# update the Dockerfile to clone the oss_fuzz branch specifically
sed 's/git clone /git clone -b oss_fuzz /' projects/xz/Dockerfile

# build project image
python3 infra/helper.py build_image 'xz'

# generate introspector report 
python3 infra/helper.py introspector 'xz'
```

JiaT75:
@mvatsyk-lsg Thanks for the commands to use for a local test. You caught my mistake but testing locally also highlighted it. I just merged your commits into master.

Great work with this! Thanks for being so flexible and responsive with all the changes.

PR #77, opened 2024-01-09
Speed up CRC32 calculation on ARM64

parheliamm:
The CRC32 instructions in ARM64 can calculate the CRC32 result for 8 bytes in a single operation, making the use of ARM64 instructions much faster compared to the general CRC32 algorithm.

Optimized CRC32 will be enabled if ARM64 has CRC extension running on Linux.

## Pull request checklist

## Pull request type

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Related Issue URL: 

## What is the new behavior?

- Enable optimized CRC32 algorithm if ARM64 support CRC extension.
  -
  -

- [ ] Yes
- [ ] No

## Other information

Benchmark data will be updated soon

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -230,6 +230,7 @@ add_library(liblzma
     src/liblzma/check/check.h
     src/liblzma/check/crc_common.h
     src/liblzma/check/crc_x86_clmul.h
+    src/liblzma/check/crc32_aarch64.h
     src/liblzma/common/block_util.c
     src/liblzma/common/common.c
     src/liblzma/common/common.h
--- a/src/liblzma/check/Makefile.inc
+++ b/src/liblzma/check/Makefile.inc
@@ -15,7 +15,8 @@ liblzma_la_SOURCES += \
 	check/check.c \
 	check/check.h \
 	check/crc_common.h \
-	check/crc_x86_clmul.h
+	check/crc_x86_clmul.h \
+	check/crc32_aarch64.h
 
 if COND_SMALL
 liblzma_la_SOURCES += check/crc32_small.c
--- a/src/liblzma/check/crc32_aarch64.h
+++ b/src/liblzma/check/crc32_aarch64.h
@@ -1,0 +1,109 @@
+///////////////////////////////////////////////////////////////////////////////
+//
+/// \file       crc32_aarch64.c
+/// \brief      CRC32 calculation with aarch64 optimization
+//
+//  Authors:    Chenxi Mao
+//
+//  This file has been put into the public domain.
+//  You can do whatever you want with this file.
+//
+///////////////////////////////////////////////////////////////////////////////
+#ifdef LZMA_CRC_CRC32_AARCH64_H
+#	error crc_arm64_clmul.h was included twice.
+#endif
+#define LZMA_CRC_CRC32_AARCH64_H
+#include <sys/auxv.h>
+// EDG-based compilers (Intel's classic compiler and compiler for E2K) can
+// define __GNUC__ but the attribute must not be used with them.
+// The new Clang-based ICX needs the attribute.
+//
+// NOTE: Build systems check for this too, keep them in sync with this.
+#if (defined(__GNUC__) || defined(__clang__)) && !defined(__EDG__)
+#	define crc_attr_target \
+        __attribute__((__target__("+crc")))
+#else
+#	define crc_attr_target
+#endif
+#ifdef BUILDING_CRC32_AARCH64
+crc_attr_target
+crc_attr_no_sanitize_address
+static uint32_t
+crc32_arch_optimized(const uint8_t *buf, size_t size, uint32_t crc)
+{
+	crc = ~crc;
+	while ((uintptr_t)(buf) & 7) {
+		crc = __builtin_aarch64_crc32b(crc, *buf);
+		buf++;
+		size--;
+	}
+	for (;size>=8;size-=8,buf+=8) {
+		crc = __builtin_aarch64_crc32x(crc, aligned_read64le(buf));
+	}
+	for (;size>0;size--,buf++)
+		crc = __builtin_aarch64_crc32b(crc, *buf);
+	return ~crc;
+}
+#endif
+#ifdef BUILDING_CRC64_AARCH64
+//FIXME: there is no crc64_arch_optimized implementation,
+// to make compiler happy, add crc64_generic here.
+#ifdef WORDS_BIGENDIAN
+#	define A1(x) ((x) >> 56)
+#else
+#	define A1 A
+#endif
+crc_attr_target
+crc_attr_no_sanitize_address
+static uint64_t
+crc64_arch_optimized(const uint8_t *buf, size_t size, uint64_t crc)
+{
+	crc = ~crc;
+
+#ifdef WORDS_BIGENDIAN
+	crc = bswap64(crc);
+#endif
+
+	if (size > 4) {
+		while ((uintptr_t)(buf) & 3) {
+			crc = lzma_crc64_table[0][*buf++ ^ A1(crc)] ^ S8(crc);
+			--size;
+		}
+
+		const uint8_t *const limit = buf + (size & ~(size_t)(3));
+		size &= (size_t)(3);
+
+		while (buf < limit) {
+#ifdef WORDS_BIGENDIAN
+			const uint32_t tmp = (uint32_t)(crc >> 32)
+					^ aligned_read32ne(buf);
+#else
+			const uint32_t tmp = (uint32_t)crc
+					^ aligned_read32ne(buf);
+#endif
+			buf += 4;
+
+			crc = lzma_crc64_table[3][A(tmp)]
+			    ^ lzma_crc64_table[2][B(tmp)]
+			    ^ S32(crc)
+			    ^ lzma_crc64_table[1][C(tmp)]
+			    ^ lzma_crc64_table[0][D(tmp)];
+		}
+	}
+
+	while (size-- != 0)
+		crc = lzma_crc64_table[0][*buf++ ^ A1(crc)] ^ S8(crc);
+
+#ifdef WORDS_BIGENDIAN
+	crc = bswap64(crc);
+#endif
+
+	return ~crc;
+}
+#endif
+static inline bool
+is_arch_extension_supported(void)
+{
+	return (getauxval(AT_HWCAP) & HWCAP_CRC32)!=0;
+}
+
--- a/src/liblzma/check/crc32_fast.c
+++ b/src/liblzma/check/crc32_fast.c
@@ -15,9 +15,12 @@
 #include "check.h"
 #include "crc_common.h"
 
-#ifdef CRC_X86_CLMUL
+#if defined(CRC_X86_CLMUL)
 #	define BUILDING_CRC32_CLMUL
 #	include "crc_x86_clmul.h"
+#elif defined(CRC32_ARM64)
+#	define BUILDING_CRC32_AARCH64
+#	include "crc32_aarch64.h"
 #endif
 
 
--- a/src/liblzma/check/crc64_fast.c
+++ b/src/liblzma/check/crc64_fast.c
@@ -14,9 +14,12 @@
 #include "check.h"
 #include "crc_common.h"
 
-#ifdef CRC_X86_CLMUL
+#if defined(CRC_X86_CLMUL)
 #	define BUILDING_CRC64_CLMUL
 #	include "crc_x86_clmul.h"
+#elif defined(CRC32_ARM64)
+#	define BUILDING_CRC64_AARCH64
+#	include "crc32_aarch64.h"
 #endif
 
 
--- a/src/liblzma/check/crc_common.h
+++ b/src/liblzma/check/crc_common.h
@@ -52,29 +52,33 @@
 #undef CRC_GENERIC
 #undef CRC_ARCH_OPTIMIZED
 #undef CRC_X86_CLMUL
+#undef CRC32_ARM64
 #undef CRC_USE_IFUNC
 #undef CRC_USE_GENERIC_FOR_SMALL_INPUTS
-
-// If CLMUL cannot be used then only the generic slice-by-eight (CRC32)
-// or slice-by-four (CRC64) is built.
-#if !defined(HAVE_USABLE_CLMUL)
-#	define CRC_GENERIC 1
 
 // If CLMUL is allowed unconditionally in the compiler options then the
 // generic version can be omitted. Note that this doesn't work with MSVC
 // as I don't know how to detect the features here.
 //
 // NOTE: Keep this this in sync with crc32_table.c.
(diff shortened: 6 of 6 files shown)


Larhzu:
Thanks! A quick few early comments (I or Jia will comment more later):

The `Data` pointer seems pointless. I don't see a reason to cast away the const qualifier.

ARM64 processors tend to support unaligned memory access but would it still be worth it (better speed) to align the `buf` first with calls to `__builtin_aarch64_crc32b`? See how aligning is done in the generic code.

I'm not sure but I guess the input `crc` variable doesn't need byte-swapping but reading the input in loop will need it. So it should use `read64le` instead, or `aligned_read64le` if `buf` is first aligned. This should be investigated or the code should be marked little-endian-only if we aren't sure.

`getauxval` is a libc function (not inline) so it's unsafe to combine it with ifunc. That is, ifunc would need to stay disabled on ARM64.

Are ARM64 processors without CRC32 common enough that runtime detection is worth it? Even they are, `#ifdef __ARM_FEATURE_CRC32` could be used at compile time to detect if CRC32 can be assumed to be supported.

The crc_edits branch is still under consideration so that may change where the code will go, possibly making things simpler.

JiaT75:
> > `getauxval` is a libc function (not inline) so it's unsafe to combine it with ifunc. That is, ifunc would need to stay disabled on ARM64.
> 
> Try to fix in the next push.
> 
> > Are ARM64 processors without CRC32 common enough that runtime detection is worth it? Even they are, `#ifdef __ARM_FEATURE_CRC32` could be used at compile time to detect if CRC32 can be assumed to be supported.
> 
> "crc" is a part of ARMv8.1 feature. To make all armv8 processors happy, it need to detect the processor feature on runtime.
> 
> > The crc_edits branch is still under consideration so that may change where the code will go, possibly making things simpler.

Hello! Thanks for the PR. The point about not needing runtime detection likely needs more research. From a quick search, my understanding is that CRC32 instruction is optional in armv8.0 and required in ARMv8.1. So if all processors in practice actually supported CRC32 in armv8.0, then it will really simplify this feature since the runtime detection adds significant complexity. The runtime detection likely requires avoiding ifunc and having different versions based on the platform (`getauxval()`, `IsProcessorFeaturePresent()`, etc).

A note on compile time detection: __ARM_FEATURE_CRC32 isn't supported by MSVC so we will need another way to detect CRC32 instruction support there.

thesamesam:
>So if all processors in practice actually supported CRC32 in armv8.0, 

Unfortunately, I have an arm64 machine without crc32, but it is an odd one.

hansjans162:
Over the holidays I got an odroid which has the crc instruction and pmull, so I have been working on an arm64 clmul implementation for crc32 and crc64. I have compared this to my implementation with clmul and found that the performance is similar to mine for smaller inputs, but becomes faster the larger the input size. From what I have found it seems like the arm64 crc instruction is also more supported than the pmull instruction that the clmul implementation depends on.

I will look into seeing how commonly supported the crc instruction and pmull are, and if runtime checks would be necessary. I have already run into cases where pmull has not been supported so it is likely we would need runtime check if we wanted to include crc64 clmul. The speed increase of crc clmul and the crc instruction were both very significant for larger bytes. The crc64 clmul reached up to 3 times as fast for these inputs.

I still need to clean up my code and remove the crc32 clmul so I can make a pr for it after this is resolved.

parheliamm:
> Over the holidays I got an odroid which has the crc instruction and pmull, so I have been working on an arm64 clmul implementation for crc32 and crc64. I have compared this to my implementation with clmul and found that the performance is similar to mine for smaller inputs, but becomes faster the larger the input size. From what I have found it seems like the arm64 crc instruction is also more supported than the pmull instruction that the clmul implementation depends on.
> 
> I will look into seeing how commonly supported the crc instruction and pmull are, and if runtime checks would be necessary. I have already run into cases where pmull has not been supported so it is likely we would need runtime check if we wanted to include crc64 clmul. The speed increase of crc clmul and the crc instruction were both very significant for larger bytes. The crc64 clmul reached up to 3 times as fast for these inputs.
> 
> I still need to clean up my code and remove the crc32 clmul so I can make a pr for it after this is resolved.

Would you please share your implementation here, I would like to double check the CRC32 speed on my side.

JiaT75:
@hansjans162 Thanks for continuing your work on ARM64 CLMUL! Your benchmarks so far sound promising and will likely be a great value to liblzma. We look forward to seeing the code and further benchmarks when they are ready :)

It sounds like runtime checks will be needed, but please let us know if anyone finds reasons to contradict this.

We did a refactor to the existing CRC related files for code organization and small optimization related reasons. This change means that architecture specific CRC optimizations should go into header files that are included in the corresponding crc32/64_fast.c file. The reasons for this are best explained in the commit message [here](https://github.com/tukaani-project/xz/commit/419f55f9dfc2df8792902b8953d50690121afeea).

Based on the refactor, @parheliamm, please put your changes into `crc32_arm64.h`, then include that header file in `crc32_fast.c`. 

Similarly, Hans please refactor what you have to create `crc64_arm64.h` with your CRC64 CLMUL work (and include this file in `crc64_fast.c`). You are likely still blocked until this PR is complete since the build changes and runtime checks should be coordinated between the branches. Feel free to send us a link to your branch whenever you have something ready that you want to show us :)

We apologize for the refactor after this was already submitted, but this PR helped inform that refactor (and gave us the needed motivation to close out the branch). Thank you everyone so far for you patience and your contributions so far!

Larhzu:
@parheliamm:
> Unaligned memory access is a default feature by ARM64, so there is no need to consider align it first.

I would expect aligned access to be faster still. In general, when unaligned access is fast on some hardware, it's fast in context of other methods for unaligned access. That is, it's much faster than doing byte-by-byte access. When the access crosses cache line or page boundaries, it may have penalties that don't occur with aligned access.

You could test with something like `uint64_t buf[2048]` and then comparing speeds of `lzma_crc32(buf, sizeof(buf) - 1, 0)` and `lzma_crc32(buf + 1, sizeof(buf) - 1, 0)`. If there is no difference then it's great news.

> I don't have big-endian test environments, so I cannot predict the behavior and unit test on big-endian.
> I think support little-endian-only is a good idea. I will modify the code.

Linux has [CRC32 assembly](https://github.com/torvalds/linux/blob/master/arch/arm64/lib/crc32.S) for both endiannesses. Maybe it can help in learning what extra steps are needed for big endian support. Having said that, little-endian-only version is fine for now, especially since it sounds that no one can test the big endian code anyway.

parheliamm:
> @parheliamm:
> 
> > Unaligned memory access is a default feature by ARM64, so there is no need to consider align it first.
> 
> I would expect aligned access to be faster still. In general, when unaligned access is fast on some hardware, it's fast in context of other methods for unaligned access. That is, it's much faster than doing byte-by-byte access. When the access crosses cache line or page boundaries, it may have penalties that don't occur with aligned access.
> 
> You could test with something like `uint64_t buf[2048]` and then comparing speeds of `lzma_crc32((unsigned char *)buf, sizeof(buf) - 1, 0)` and `lzma_crc32((unsigned char *)buf + 1, sizeof(buf) - 1, 0)`. If there is no difference then it's great news.
> 
> > I don't have big-endian test environments, so I cannot predict the behavior and unit test on big-endian.
> > I think support little-endian-only is a good idea. I will modify the code.
> 
> Linux has [CRC32 assembly](https://github.com/torvalds/linux/blob/master/arch/arm64/lib/crc32.S) for both endiannesses. Maybe it can help in learning what extra steps are needed for big endian support. Having said that, little-endian-only version is fine for now, especially since it sounds that no one can test the big endian code anyway.
> 
> **Edited** to add missing casts to the `lzma_crc32` calls.

Based on my test, there is no difference between aligned access and unaligned access on my kunpeng 920 ARM64 platform.
Here is the test result:
lge@kunpeng920:~/SATA/work/CRC$ ./crc
algined，time: 1454.59 ms, address=0xcfad0060
crc=0xb596e05e
unalgined，time: 1454.59 ms, address=0xd0ad0061
crc=0xb596e05e

parheliamm:
> @parheliamm:
> 
> > Unaligned memory access is a default feature by ARM64, so there is no need to consider align it first.
> 
> I would expect aligned access to be faster still. In general, when unaligned access is fast on some hardware, it's fast in context of other methods for unaligned access. That is, it's much faster than doing byte-by-byte access. When the access crosses cache line or page boundaries, it may have penalties that don't occur with aligned access.
> 
> You could test with something like `uint64_t buf[2048]` and then comparing speeds of `lzma_crc32((unsigned char *)buf, sizeof(buf) - 1, 0)` and `lzma_crc32((unsigned char *)buf + 1, sizeof(buf) - 1, 0)`. If there is no difference then it's great news.
> 
> > I don't have big-endian test environments, so I cannot predict the behavior and unit test on big-endian.
> > I think support little-endian-only is a good idea. I will modify the code.
> 
> Linux has [CRC32 assembly](https://github.com/torvalds/linux/blob/master/arch/arm64/lib/crc32.S) for both endiannesses. Maybe it can help in learning what extra steps are needed for big endian support. Having said that, little-endian-only version is fine for now, especially since it sounds that no one can test the big endian code anyway.
> 
> **Edited** to add missing casts to the `lzma_crc32` calls.

Based on my local test, unaligned access is slower than aligned access.
So I will modify the code to handle unaligned data first.

parheliamm:
I've updated aarch64 CRC code with align memory access.
Other parts will be rebased to the latest code.

hansjans162:
> Would you please share your implementation here, I would like to double check the CRC32 speed on my side. For CRC32d insns, I can unroll it for larger size, e.g 4K or 16K like google did, that will be much better performance but larger code size.

@parheliamm [Here](https://github.com/hansjans162/xz/tree/tmp-arm64-crc-clmul) is the arm64 clmul implementation. Sorry for the delay I was updating my code to use the new organization. I would also like to note that my code is not finished yet. is_arch_extension_supported currently always returns true, and platform checks in configure.ac and CMakeLists.txt need to be made.

parheliamm:
> I've updated aarch64 CRC code with align memory access. Other parts will be rebased to the latest code.

@hansjans162 :
I did a quick tests for CRC32s with 16MB data with 1024 loops:
```
lge@kunpeng920:~/SATA/work/CRC$ ./crc
algined crc32d，time: 1268.48 ms, address=0xc52d0060
crc=0xb596e05e
unalgined crc32d，time: 1264.09 ms, address=0xc62d0061
crc=0xb596e05e
unaligned crc32_arch_optimized，time: 3178.84 ms, address=0xc62d0061
crc=0xb596e05e
```
The crc32_arch_optimized seems takes more time on my local test.

hansjans162:
> The crc32_arch_optimized seems takes more time on my local test.

@parheliamm I'm glad your test result match up with mine. I still found that the arm64 clmul implementation for crc64 is faster than the generic and worth keeping. Maybe you can test this to confirm? I will remove crc32 clmul from my code, but keep in crc64 clmul.

parheliamm:
> > The crc32_arch_optimized seems takes more time on my local test.
> 
> @parheliamm I'm glad your test result match up with mine. I still found that the arm64 clmul implementation for crc64 is faster than the generic and worth keeping. Maybe you can test this to confirm? I will remove crc32 clmul from my code, but keep in crc64 clmul.

@hansjans162 
I will check CRC64 with clmul on my side.
I believed the CRC64 clmul is faster than generic code.
I will share the result later.

JiaT75:
@parheliamm @hansjans162 Thank you for your patience on this review. We had a few other things we were focusing on (new website, release, etc.). I promise I did not forget about this :)

I created a [branch](https://github.com/tukaani-project/xz/tree/arm64_crc32) with some additions to this PR that will also be helpful for the CRC64 CLMUL. The extent of the edits should be clear from the commit messages and code changes, but let me know if you have questions about the changes.

Can you both test this on your hardware to be sure it works correctly (and is still fast)? I was able to cross-compile it with GCC, Clang, and MSVC but I do not have an ARM64 device so I have not run the code. Thanks!

hansjans162:
> @parheliamm @hansjans162 Thank you for your patience on this review. We had a few other things we were focusing on (new website, release, etc.). I promise I did not forget about this :)
> 
> I created a [branch](https://github.com/tukaani-project/xz/tree/arm64_crc32) with some additions to this PR that will also be helpful for the CRC64 CLMUL. The extent of the edits should be clear from the commit messages and code changes, but let me know if you have questions about the changes.
> 
> Can you both test this on your hardware to be sure it works correctly (and is still fast)? I was able to cross-compile it with GCC, Clang, and MSVC but I do not have an ARM64 device so I have not run the code. Thanks!

I tested your code and found a problem with the crc32_arch_optimized function. the updated function below should fix this.

I made two changes to this function. First was making align_amount the difference from 8 instead of just the remainder. I also xor this with 8 so buf_end does not change if it is already properly aligned. The second thing was changing __crc32w to __crc32d.
```c
crc_attr_target
static uint32_t
crc32_arch_optimized(const uint8_t *buf, size_t size, uint32_t crc)
{
	crc = ~crc;

	// Align the input buffer because this was shown to be
	// significantly faster than unaligned accesses.
	const size_t align_amount = my_min(size, ((8-((uintptr_t)(buf) & 7))^8));
	const uint8_t *buf_end = buf + align_amount;

	for (; buf < buf_end; ++buf)
		crc = __crc32b(crc, *buf);

	for (buf_end = ((size - align_amount) - 8) + buf; buf < buf_end;
			buf += 8)
		crc = __crc32d(crc, aligned_read64le(buf));

	for (buf_end += 8; buf < buf_end; ++buf)
		crc = __crc32b(crc, *buf);

	return ~crc;
}
```

Larhzu:
```
for (buf_end = ((size - align_amount) - 8) + buf; buf < buf_end;
		buf += 8)

```

I suspect that `clang -fsanitize=undefined` will complain at runtime. If `size` equals 1 and `align_amount` equals 0 or 1, it ends up calculating `buf - 8` or `buf - 7`. That is, the pointer arithmetic may go beyond the beginning of the buffer, which the C standard doesn't allow (but one element past the end is allowed).

Although it should work in practice on ARM64 (unless the buffer is at a weird address where the address would overflow but that's unlikely), I think it should be possible to avoid the problem without a performance penalty.

JiaT75:
> ```
> for (buf_end = ((size - align_amount) - 8) + buf; buf < buf_end;
> 		buf += 8)
> ```
> 
> I suspect that `clang -fsanitize=undefined` will complain at runtime. If `size` equals 1 and `align_amount` equals 0 or 1, it ends up calculating `buf - 8` or `buf - 7`. That is, the pointer arithmetic may go beyond the beginning of the buffer, which the C standard doesn't allow (but one element past the end is allowed).
> 
> Although it should work in practice on ARM64 (unless the buffer is at a weird address where the address would overflow but that's unlikely), I think it should be possible to avoid the problem without a performance penalty.

Thanks for pointing this out! I pushed a new version to avoid pointer arithmetic beyond the beginning of the buffer. I also added macOS support, but I couldn't test it since I do not have an Apple device. The Apple specific code is small so it should work but I hope someone can test this.

PR #108, opened 2024-04-13
CI: Use only the active CPUs on macOS

ryandesign:
`hw.ncpu` counts all CPUs including inactive ones. `hw.activecpu` counts only the active CPUs.


thesamesam (approved):
Thank you!

shighetari:
Much appreciated!

PR #111, opened 2024-04-17
New XZ Logo proposal

Sur3:
I saw the old 'tainted' Logo for XZ was removed; here is a new Logo proposal in pixelart style combining the letters X and Z in a minimalistic color-coding way in pixel and 8x resolution. If you like this idea I might do an SVG version aswell. As for License do what you want, as by German law I would still own the copyright, but the Logo is minimalistic enough that I would assume it any License you as Maintainers want.


AlejandroBarreda:
nice

Larhzu:
Thanks!

I'm happy with _Bob the Toucan_ as the Tukaani project logo. I think the XZ projects don't need a logo. Jia wanted one and I wasn't really _against_ it. That logo is now gone.

So, thank you for the effort but I don't plan to have a new logo.

PR #116, opened 2024-04-25
CI: actually fail on FreeBSD

junghans:
Without "set -e" the job will always be successful.

See vmactions/freebsd-vm#72

/cc @thesamesam


junghans:
We might have to do the same on Net/OpenBSD.

PR #117, opened 2024-05-01
Update SECURITY.md

JohnDoe1001:
Added the requirement to include security info, reproduction steps, estimated severity and encouraging people to report the issue publicly.

### DIFF
--- a/.github/SECURITY.md
+++ b/.github/SECURITY.md
@@ -1,13 +1,21 @@
 # Security Policy
 
-If you discover a security vulnerability in this project, please
-report it privately. **Do not disclose it as a public issue.**
+If you discover a security vulnerability in this project, especially a highly dangerous one, you can either report it privately or disclose it as a public issue, so that more people are informed of the situation.
+
+In any case, please
+ provide a clear description of the vulnerability including:
+
+ - Affected versions of XZ Utils
+ - Estimated severity (low, moderate, high, critical)
+ - Steps to recreate the vulnerability
+ - All relevant files (core dumps, build logs, input files, etc....
+
 
 You may submit a report via email to
-[Lasse Collin](mailto:lasse.collin@tukaani.org)
-(OpenPGP key fingerprint: 3690 C240 CE51 B467 0D30 AD1C 38EE 757D 6918 4620),
-or through
-[Security Advisories](https://github.com/tukaani-project/xz/security/advisories/new).
+- [xz@tukaani.org](mailto:xz@tukaani.org)
+- [Lasse Collin](mailto:lasse.collin@tukaani.org)
+(OpenPGP key fingerprint: 3690 C240 CE51 B467 0D30 AD1C 38EE 757D 6918 4620)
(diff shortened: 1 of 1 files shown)


JohnDoe1001:
I think that every security exploits should be reported publicly and clearly, so that the community & developers could fix it ASAP.

What’s your opinion?

.github/SECURITY.md
```diff
@@ -1,13 +1,21 @@
 # Security Policy
 
-If you discover a security vulnerability in this project, please
-report it privately. **Do not disclose it as a public issue.**
+If you discover a security vulnerability in this project, especially a highly dangerous one, you can either report it privately or disclose it as a public issue, so that more people are informed of the situation.
```
eli-schwartz, line 3: This goes against literally all security best practices, which leaves the question: why are you suggesting this?

eli-schwartz (changes requested):
> I think that every security exploits should be reported publicly and clearly, so that the community & developers could fix it ASAP.

I'm not sure how to respond other than "no".

.github/SECURITY.md
```diff
@@ -1,13 +1,21 @@
...
+ - All relevant files (core dumps, build logs, input files, etc....
+
 
 You may submit a report via email to
-[Lasse Collin](mailto:lasse.collin@tukaani.org)
-(OpenPGP key fingerprint: 3690 C240 CE51 B467 0D30 AD1C 38EE 757D 6918 4620),
-or through
-[Security Advisories](https://github.com/tukaani-project/xz/security/advisories/new).
+- [xz@tukaani.org](mailto:xz@tukaani.org)
+- [Lasse Collin](mailto:lasse.collin@tukaani.org)
```
eli-schwartz, line 16: Why is the xz@ address being added back here?

eli-schwartz (changes requested):
The commit message "Update SECURITY.md" is inherently problematic, as it doesn't describe what the proposed change actually does.

Nowhere in the commit message does it clarify why it makes the changes it does -- in some cases, it doesn't even mention that the changes are being made, even though they seem... important?

Larhzu:
Fast full disclosure was the only sensible thing with the backdoor incident. With more typical security issues it seems sensible to try to prepare a fix first so that fewer bad actors will have time to use the information. So one has to use some common sense when one finds a security issue.

I assume that people are capable of providing the required information without detailed handholding in a form of a bullet point list. I asked Jia to simplify the file and also the issue and PR templates and I don't want to restore such cruft. If something is missing from a report, it's possible to ask for more information.

It seems that `SECURITY.md` still keeps attracting posts and this makes me wonder if the existence of that file really does more good than bad. People had no trouble reporting issues via email before the project was on GitHub. (Or maybe email should be the only method listed in `SECURITY.md` as it's simpler than the GH feature.)

For example, CVE-2022-1271 (ZDI-CAN-16587) was reported to me and then I reported it to GNU gzip maintainers since the bug was inherited to `xzgrep` from `zgrep`. It took several emails to polish the fix and coordinate the disclosure with xz, gzip, and downstream distros (distros via mailing list on openwall.com). I think the process worked really well.

PR #118, opened 2024-05-01
Mask control characters in filenames

Larhzu:
The command line tools print filenames and some other user-specified strings to standard output or standard error. Malicious strings, for example, from filenames could contain control characters that affect the state of the terminal.

These commits add a function to replace the single-byte control characters with question marks. This is simple but hopefully good enough in practice.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1649,6 +1649,8 @@ if(NOT MSVC OR MSVC_VERSION GREATER_EQUAL 1900)
         src/common/sysdefs.h
         src/common/tuklib_common.h
         src/common/tuklib_config.h
+        src/common/tuklib_mbstr_nonprint.c
+        src/common/tuklib_mbstr_nonprint.h
         src/common/tuklib_exit.c
         src/common/tuklib_exit.h
         src/common/tuklib_gettext.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1595,6 +1595,8 @@ if(HAVE_DECODERS AND (NOT MSVC OR MSVC_VERSION GREATER_EQUAL 1900))
         src/common/sysdefs.h
         src/common/tuklib_common.h
         src/common/tuklib_config.h
+        src/common/tuklib_mbstr_nonprint.c
+        src/common/tuklib_mbstr_nonprint.h
         src/common/tuklib_exit.c
         src/common/tuklib_exit.h
         src/common/tuklib_gettext.h
@@ -1619,6 +1621,7 @@ if(HAVE_DECODERS AND (NOT MSVC OR MSVC_VERSION GREATER_EQUAL 1900))
     endif()
 
     tuklib_progname(lzmainfo)
+    tuklib_mbstr(lzmainfo)
 
     # NOTE: The translations are in the "xz" domain and the .mo files are
     # installed as part of the "xz" target.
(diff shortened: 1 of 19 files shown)


Larhzu:
Single-byte control character masking isn't enough. At least Konsole and Xfce Terminal (but not uxterm) interpret C1 control codes and CSI sequences in `en_US.UTF-8` locale.

```
$ printf 'foo\u009b3Dbar\n'
bar

$ printf 'a\u0090 Can you see me? \u009cb\n'
ab
```

A proper masking method must decode multibyte characters. It must tolerate invalid multibyte sequences and restart decoding from the next byte.

src/common/tuklib_cntrl_chars.c
```diff
@@ -0,0 +1,59 @@
...
+			return true;
+
+	return false;
+}
+
+
+extern const char *
+tuklib_mask_cntrl(const char *str)
+{
+	static char *mem = NULL;
```
rcolyer, line 37: Threading is utilized, so this src/common function should probably have static thread_local char *mem to avoid future issues.

Larhzu: `thread_local` needs C11 but the codebase is C99 for now. The function is documented to be thread unsafe. It's called from the main thread only.

The function is really primitive as it doesn't even provide more than one memory slot. That is, one cannot print a message that needs two masked strings. It's easy to expand but the first version doesn't need to do more than is required right now.

src/common/tuklib_cntrl_chars.h
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: 0BSD
+
+///////////////////////////////////////////////////////////////////////////////
+//
+/// \file       tuklib_cntrl_chars.h
+/// \brief      Find and replace single-byte control characters
+///
+/// This is a small and very simple implementation that uses isnctrl(3),
```
jer-gentoo, line 8: `isnctrl` => `iscntrl`

Larhzu:
The new version should handle all _relevant_ multibyte character sets, not just UTF-8.

Instead of looking for control characters, it now looks for non-printable characters which is a much stricter check. A possible downside is that an old C library might not recognize newer printable Unicode characters even though the user might be using them already. I suppose it's not a real problem. :-) Gnulib's `quotearg` looks for printable characters as well.

`tuklib_mask_nonprint` isn't thread safe although it would be straightforward to add a variant that takes a `char **mem` argument. That wouldn't require C11's _optional feature_ `thread_local` from `<threads.h>` or a compiler-specific thread-local extension (although those are quite widely available). Then it would be possible to have more than one masked string available at the same time in case one needs to print two filenames at once. But the simplest version is enough for now.

I hope I didn't miss any string that should be masked.

Larhzu:
I merged this with a few changes:

* One filename usage wasn't masked in `list.c`.

* Don't reject non-printable characters in `xz --suffix=.SOME_BAD_CHARS`. It would have been problematic especially if `LC_CTYPE` is `C` and one tries to use fancy chars in a suffix. Only mask the chars when printing the suffix or derived filename in messages.

* I added a note to a [commit message](https://github.com/tukaani-project/xz/commit/d22f96921fd2f94d842f3cc2e5f729cb3cca5122) about `xz --robot --list`.

* [Handle U+FFFD specially on Windows](https://github.com/tukaani-project/xz/blob/16c9796ef970ae349c54fef9a346e394d7cc4c75/src/common/tuklib_mbstr_nonprint.c#L57).

PR #120, opened 2024-05-02
WIP: meson build

vtorri:


src/liblzma/meson.build
```diff
@@ -0,0 +1,360 @@
...
+pkg_mod.generate(
+    lzma_lib,
+    description : 'General purpose data compression library',
+    name : 'liblzma',
+    url : 'https://tukaani.org/xz/',
+)
+
+install_headers('api/lzma.h',
+    install_dir: get_option('prefix') / get_option('includedir')
+)
```
eli-schwartz, line 339: The default install dir is already the includedir, no need to override it.

src/liblzma/meson.build
```diff
@@ -0,0 +1,360 @@
...
+    'api/lzma/index_hash.h',
+    'api/lzma/lzma12.h',
+    'api/lzma/stream_flags.h',
+    'api/lzma/version.h',
+    'api/lzma/vli.h',
+])
+
+install_headers(lzma_headers,
+    install_dir: get_option('prefix') / get_option('includedir') / 'lzma'
+)
```
eli-schwartz, line 360:
```suggestion
install_headers(lzma_headers,
    subdir: 'lzma',
)
```

eli-schwartz (changes requested):
At a quick glance, I have a couple suggestions.

src/liblzma/meson.build
```diff
@@ -0,0 +1,358 @@
...
+    lzma_deps += pthread_dep
+endif
+
+lzma_lib = library('lzma', lzma_src,
+    c_args : lzma_c_args,
+    dependencies : lzma_deps,
+    gnu_symbol_visibility : 'hidden',
+    include_directories : lzma_incdir,
+    install : true,
+    link_args : sys_windows ? '-Wl,--output-def,liblzma.def.in' : '',
```
Larhzu, line 322: `--output-def` is for GNU `ld` and perhaps also `lld` from LLVM. CMakeLists.txt uses this option behind `if(NOT MSVC)`. With CMake, the package builds with both GCC and Clang/LLVM + MinGW-w64.

Larhzu:
The following is relevant only when other things have already been finished. I'm writing it here so that it won't be forgotten.

The tuklib files are a bit similar to Gnulib in sense that a subset of the files is meant to be reusable in other projects somewhat easily. With Autotools and CMake, one can copy the relevant `m4/tuklib_*.m4` or `cmake/tuklib_*.cmake` files along with the `.h` and `.c` files.

Meson's language is intentionally more restrictive so it's OK that the first iterations put all tuklib checks into a single `meson.build`. This is not ideal in the long term though as it makes it difficult to reuse a subset of the modules.

One method would be to have a separate repository for the tuklib modules, each module having a `meson.build` snippet file. Then a script would take a list of modules to use and generate a single `meson.build` from the snippets (in the right order, handling possible dependencies). The downside is that each project using a different subset of the modules would end up with _a different generated build system file_ in the source package (simpler than `configure` but still).

Another idea could be to have a variable, for example, `tuklib_modules` which would contain a list of modules to use. Top-level `meson.build` would set that variable and then do `subdir(src/common/tuklib_meson)` which would further do `subdir(modulename123)` for each module name that was found in `tuklib_modules` (in dependency order).

This way the checks of each module would be in their own `meson.build` files. In the parent directory the `meson.build` file would list all modules, including those that a specific project doesn't need. It would be just a boring chain of if-subdir-endif lines in the correct order to satisfy possible cross-module dependencies. This way no generated files would be needed to use a subset of the modules in another project.

(The if-subdir-endif file might still be a generated file in the module repository. However, it would be the same for all projects that use the modules, thus there wouldn't be as many versions of the generated file, making reviews easier.)

The set of tuklib modules is tiny so in this sense this is silly overthinking. But I'm curious how this would have to be done with bigger number of reusable modules (like Gnulib). It seems essential to split the per-module checks into separate files if there are tens of modules and different projects use a different set of modules. A per-project generated `meson.build` that contains checks for tens of modules sounds too similar to a generated `configure`. Avoiding such per-project generated files is a big reason why this Meson discussion in XZ Utils is happening in the first place.

src/liblzma/meson.build
vtorri: yes, I know, that's something that will be fixed.

meson.build
```diff
@@ -0,0 +1,376 @@
...
+    license_files: [
+        'COPYING',
+        'COPYING.0BSD',
+        'COPYING.GPLv2',
+        'COPYING.GPLv3',
+        'COPYING.LGPLv2.1'
+    ],
+    default_options: [
+        'b_ndebug=if-release',
+        'c_std=c11',
```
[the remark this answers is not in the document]
vtorri: I know that some project leaders do not like to use the latest version of meson. If there is no problem for you to use meson >= 1.3, then ok.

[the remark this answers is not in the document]
Larhzu: I actually don't know right now which Meson (or muon) version is OK to require. I just get an impression that a single value doesn't work everywhere (if MSVC needs C11 mode). I haven't checked in detail how the `c_std` option in Meson exactly works.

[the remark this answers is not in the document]
eli-schwartz: https://github.com/mesonbuild/meson/blob/master/mesonbuild/compilers/c.py#L453-L485

For MSVC, meson ignores c99 and passes /std:XXX when the mode is c11 or c17. The difference is really about whether you want to opt into GNU extensions (which MSVC obviously doesn't support).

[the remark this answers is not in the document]
Larhzu: I vaguely remember that, a very long time ago, using `gnu99` was better in practice when the intention was to use non-standard extensions (language or library) when they are available.

If `gnu99` doesn't put MSVC into C11/C17 mode then MSVC needs some other `c_std` option. MSVC is a special case in so many ways that handling it specially here could be OK too, assuming it's easy to do.

Using `gnu11` with GCC and Clang and strict C11 with compilers that don't support GNU extensions could be OK too. But if compiler doesn't support C11 then it should fallback to C99 if that is supported.

The list method that requires Meson >= 1.3.0 would make all this trivially easy. [muon](https://muon.build/releases/edge/docs/status.html) says it's currently at _meson compat version 1.1_.

So, oh well, maybe try `gnu11` for now if that helps getting MSVC to C11 mode. This is a small detail in the big picture and can be worried more a little later.

meson.build
```diff
@@ -0,0 +1,376 @@
...
+    endif
+
+    foreach m : match_finders
+        if get_option('match-finders').contains(m) == true
+            config_h.set10('HAVE_MF_' + m.underscorify().to_upper(), true,
+                description: 'Define to 1 to enable ' + m + ' match finder.'
+            )
+            match_finder_msg += m + ' '
+        endif
+    endforeach
```
nirbheek, line 259:
```suggestion
    foreach m : get_option('match-finders')
        config_h.set10('HAVE_MF_' + m.underscorify().to_upper(), true,
            description: 'Define to 1 to enable ' + m + ' match finder.'
        )
        match_finder_msg += m + ' '
    endforeach
```

Then you can get rid of the duplicated `match_finders` above, and keep it in one place: `meson_options.txt`. Meson will never let you set a value to this option that's not in the allowed options list.

meson.build
```diff
@@ -0,0 +1,376 @@
...
+]
+
+foreach c : checks
+    if get_option('checks').contains(c) == true
+        config_h.set10('HAVE_CHECK_' + c.underscorify().to_upper(), true,
+            description: 'Define to 1 if ' + c + ' integrity check is enabled.'
+        )
+        check_msg += c + ' '
+    endif
+endforeach
```
nirbheek, line 279:
```suggestion
foreach c : get_option('checks')
    config_h.set10('HAVE_CHECK_' + c.underscorify().to_upper(), true,
        description: 'Define to 1 if ' + c + ' integrity check is enabled.'
    )
    check_msg += c + ' '
endforeach
```

meson.build
```diff
@@ -0,0 +1,376 @@
...
+
+if get_option('checks').contains('crc32') == false
+    error('For now, the CRC32 check must always be enabled')
+endif
+
+### microLZMA
+
+if get_option('microlzma') == true
```
nirbheek, line 287:
```suggestion
if get_option('microlzma')
```

All `== true` comparisons are redundant.

There are dozens of instances of this sort of issue. This is a really weird pattern that difficult to read, you're basically doing true == true in every case, because Meson does not allow type coercion to `bool` (not even from `int` to `bool`), nor does it allow comparisons between different types.

This isn't C :wink:

src/common/meson.build
```diff
@@ -0,0 +1,400 @@
...
+                res = cc.compiles(ua_src, name: 'fast unaligned access for ' + cpu)
+                if res == true
+                    config_h.set10('TUKLIB_FAST_UNALIGNED_ACCESS', true,
+                    description: 'Define to 1 if the system supports fast unaligned access to 16-bit, 32-bit, and 64-bit integers.'
+                    )
+                    fast_unaligned_access = true
+                    break
+                endif
+            endif
+        endforeach
```
nirbheek, line 222:
```suggestion
        if host_machine.cpu_family() in unaligned_access_cpus
            ua_src = '''
#if !defined(__ARM_FEATURE_UNALIGNED) \
		&& !defined(__riscv_misaligned_fast) \
		&& !defined(_MSC_VER)
compile error
#endif
int main(void) { return 0; }
'''
            if cc.compiles(ua_src, name: 'fast unaligned access for ' + cpu)
                config_h.set10('TUKLIB_FAST_UNALIGNED_ACCESS', true,
                description: 'Define to 1 if the system supports fast unaligned access to 16-bit, 32-bit, and 64-bit integers.'
                )
                fast_unaligned_access = true
            endif
        endif
```

src/common/meson.build
```diff
@@ -0,0 +1,400 @@
...
+
+    if res == true
+        if cc.has_function('mbrtowc',
+            prefix : '#include <wchar.h>'
+            )
+            config_h.set10('HAVE_MBRTOWC', true,
+                description: 'Define to 1 if the mbrtowc() function is available.'
+            )
+        endif
+    endif
```
nirbheek, line 257:
```suggestion
    if res and cc.has_function('mbrtowc', prefix : '#include <wchar.h>')
        config_h.set10('HAVE_MBRTOWC', true,
            description: 'Define to 1 if the mbrtowc() function is available.'
        )
    endif
```

src/liblzma/meson.build
```diff
@@ -0,0 +1,358 @@
...
+    endif
+
+    if get_option('decoders').contains('lzma2') == true
+        lzma_src += [
+            'lzma/lzma2_decoder.c'
+        ]
+    endif
+
+    # rangecoder/Makefile.inc
+    if get_option('encoders').contains('lzma1') == true
```
nirbheek, line 221: This is the same condition as above that adds `'lzma/lzma_encoder.c'` ?

src/liblzma/meson.build
```diff
@@ -0,0 +1,358 @@
...
+    endif
+endif
+
+if get_option('encoders').contains('lzma1') == true or
+   get_option('decoders').contains('lzma1') == true
+
+    # lzma/Makefile.inc
+    lzma_src += [
+        'lzma/lzma_encoder_presets.c'
+    ]
```
nirbheek, line 186: Only this line needs to be inside this `if`, the rest of the lines can be outside since their comparisons are a subset of this `if`.

src/liblzma/meson.build
```diff
@@ -0,0 +1,358 @@
...
+        ]
+    endif
+
+    if get_option('decoders').contains('lzma1') == true or
+       get_option('decoders').contains('lzma2') == true
+        lzma_src += [
+            'lz/lz_decoder.c'
+        ]
+    endif
+endif
```
nirbheek, line 178:
```suggestion
# lz/Makefile.inc
if 'lzma1' in get_option('encoders') or 'lzma2' in get_option('encoders')
    lzma_src += [
        'lz/lz_encoder.c',
        'lz/lz_encoder_mf.c'
    ]
endif

if 'lzma1' in get_option('decoders') or 'lzma2' in get_option('decoders')
    lzma_src += [
        'lz/lz_decoder.c'
    ]
endif
```

The outer 'if' is redundant.

src/liblzma/meson.build
```diff
@@ -0,0 +1,358 @@
...
+
+if get_option('encoders').contains('delta') == true or
+   get_option('decoders').contains('delta') == true
+
+    # delta/Makefile.inc
+    lzma_src += [
+        'delta/delta_common.c'
+    ]
```
nirbheek, line 234: Same, only this line needs to be inside this if, the rest of the lines can be outside since their comparisons are a subset of this if.

src/liblzma/meson.build
```diff
@@ -0,0 +1,358 @@
...
+        lzma_src += [
+            'simple/simple_encoder.c'
+        ]
+    endif
+
+    if have_decoder_simple_filters == true
+        lzma_src += [
+            'simple/simple_decoder.c'
+        ]
+    endif
```
nirbheek, line 266: These `if` statements do not have to nest inside this `if` statement

meson.build
```diff
@@ -0,0 +1,376 @@
...
+        )
+        encoder_msg += f + ' '
+    endif
+    if get_option('decoders').contains(f)
+        config_h.set10('HAVE_DECODER_' + f.underscorify().to_upper(), true,
+            description: 'Define to 1 if ' + f + ' decoder is enabled.'
+        )
+        decoder_msg += f + ' '
+    endif
+endforeach
```
nirbheek, line 206: It looks like `supported_filters` is just duplicating the list of `encoders` and `decoders` defined in `meson_options.txt`, so we can simplify to:

```suggestion
simple_filters = [
    'x86',
    'powerpc',
    'ia64',
    'arm',
    'armthumb',
    'arm64',
    'sparc',
    'riscv',
]

# FIXME foreach i in array; set_variable(f'have_@i@', true) endforeach

foreach f : get_option('encoders')
    config_h.set10('HAVE_ENCODER_' + f.underscorify().to_upper(), true,
        description: 'Define to 1 if ' + f + ' encoder is enabled.'
    )
    encoder_msg += f + ' '
endforeach
foreach f : get_option('decoders')
    config_h.set10('HAVE_DECODER_' + f.underscorify().to_upper(), true,
        description: 'Define to 1 if ' + f + ' decoder is enabled.'
    )
    decoder_msg += f + ' '
endforeach
```

meson.build
```diff
@@ -0,0 +1,382 @@
...
+        'COPYING.0BSD',
+        'COPYING.GPLv2',
+        'COPYING.GPLv3',
+        'COPYING.LGPLv2.1'
+    ],
+    default_options: [
+        'b_ndebug=if-release',
+        'buildtype=debug',
+        'c_std=c99',
+        'libdir=lib',
```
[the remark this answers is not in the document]
vtorri: done in fc52493cced9470318026574d94420c86227bcb7

meson.build
```diff
@@ -0,0 +1,382 @@
...
+# try mimic AC_USE_SYSTEM_EXTENSIONS
+if host_machine.system() == 'linux'
+    add_global_arguments('-D_GNU_SOURCE', language: 'c')
+elif host_machine.system() == 'sunos'
+    add_global_arguments('-D__EXTENSIONS__', language: 'c')
+    add_global_arguments('-D_POSIX_PTHREAD_SEMANTICS', language: 'c')
+else
+    add_global_arguments('-D_TANDEM_SOURCE', language: 'c')
+    add_global_arguments('-D_ALL_SOURCE', language: 'c')
+endif
```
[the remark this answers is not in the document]
vtorri: done in fc52493cced9470318026574d94420c86227bcb7

meson.build
```diff
@@ -0,0 +1,382 @@
...
+        '-Wnewline-eof',
+        '-Wshift-sign-overflow',
+        '-Wstring-conversion',
+    ]
+
+    foreach cf : cflags_try
+        if cc.has_argument(cf)
+            lzma_c_args += cf
+        endif
+    endforeach
```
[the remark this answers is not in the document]
vtorri: done in fc52493cced9470318026574d94420c86227bcb7

src/common/meson.build
```diff
@@ -0,0 +1,400 @@
...
+int main()
+{
+	__builtin_bswap16(1);
+	__builtin_bswap32(1);
+	__builtin_bswap64(1);
+	return 0;
+}
+'''
+res = cc.links(bswap_src, name: '__builtin_bswap16/32/64')
+if res == true
```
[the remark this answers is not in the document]
vtorri: done in 72f696e5426cfa3e6ee9e294108e051d39890253

meson.build
```diff
@@ -0,0 +1,382 @@
...
+foreach f : simple_filters
+    if get_option('encoders').contains(f) == true
+        have_encoder_simple_filters = true
+    endif
+    if get_option('decoders').contains(f) == true
+        have_decoder_simple_filters = true
+    endif
+endforeach
+
+if (get_option('encoders').contains('lzma2') == true and get_option('encoders').contains('lzma1') == false) or (get_option('decoders').contains('lzma2') == true and get_option('decoders').contains('lzma1') == false)
```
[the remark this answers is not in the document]
vtorri: done in 72f696e5426cfa3e6ee9e294108e051d39890253

src/liblzma/meson.build
vtorri: done in fc52493cced9470318026574d94420c86227bcb7

src/liblzma/meson.build
vtorri: done in fc52493cced9470318026574d94420c86227bcb7

meson.build
```diff
@@ -0,0 +1,382 @@
...
+    'armthumb',
+    'arm64',
+    'sparc',
+    'riscv',
+]
+
+# FIXME foreach i in array; set_variable(f'have_@i@', true) endforeach
+
+foreach f : supported_filters
+    if get_option('encoders').contains(f) == true
```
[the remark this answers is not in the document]
vtorri: done in fc52493cced9470318026574d94420c86227bcb7

meson.build
```diff
@@ -0,0 +1,376 @@
...
+
+match_finders = [
+    'hc3',
+    'hc4',
+    'bt2',
+    'bt3',
+    'bt4'
+]
+
+if get_option('encoders').contains('lzma2') == true or get_option('encoders').contains('lzma1') == true
```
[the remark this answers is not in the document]
vtorri: done in 72f696e5426cfa3e6ee9e294108e051d39890253

src/common/meson.build
```diff
@@ -0,0 +1,400 @@
...
+        if host_machine.cpu_family() == cpu
+            config_h.set10('TUKLIB_FAST_UNALIGNED_ACCESS', true,
+                description: 'Define to 1 if the system supports fast unaligned access to 16-bit, 32-bit, and 64-bit integers.'
+            )
+            fast_unaligned_access = true
+            break
+        endif
+    endforeach
+
+    if fast_unaligned_access == false
```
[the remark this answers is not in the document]
vtorri: done in 72f696e5426cfa3e6ee9e294108e051d39890253

meson.build
```diff
@@ -0,0 +1,376 @@
+## SPDX-License-Identifier: 0BSD
+## Author: Vincent Torri
+
+project(
+    'XZ Utils',
```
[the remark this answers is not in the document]
vtorri: meson should add a 'distname' option to project()

meson.build
vtorri: i'm too used to C indeed :-)
done in 72f696e5426cfa3e6ee9e294108e051d39890253

meson.build
vtorri: done in 7056bbc2f62cb4c6e045ec93bcaca0faa9451dde

src/liblzma/meson.build
thesamesam: Please add a comment indicating that. It's hard to know if something is wrong by omission or as a TODO.

meson.build
vtorri: done in d8238215f5cdb8225ab4660d1f63c66bf76b59d5

src/common/meson.build
vtorri: done in 8f0731e245f0af8fd023877f774006d4346cc2fa

src/liblzma/meson.build
vtorri: fixed in 3f290102fa92bb1dbd08402979902499687b2b7c

src/liblzma/meson.build
```diff
@@ -155,26 +155,18 @@ if ('sha256' in get_option('checks')) and (not get_option('external-sha256'))
...
-            'lz/lz_decoder.c'
-        ]
-    endif
+if ('lzma1' in get_option('decoders')) or ('lzma2' in get_option('decoders'))
+    lzma_src += [
+        'lz/lz_decoder.c'
+    ]
 endif
 
 if ('lzma1' in get_option('encoders')) or ('lzma1' in get_option('decoders'))
```
thesamesam, line 172: I wonder if things like this should be factored out into their own variable, like:
```meson
have_lzma1_encoder = ...
```

vtorri: i wonder if lzma1 and lzma2 should be not an option, but just required, as it's the main purpose of xz. That would simplify the build a LOT...

src/liblzma/meson.build
vtorri: not so simple: lzma/Makefile.inc is included if lzma1 is enabled, but in lzma/Makefile.inc there are some code if lzma2 is enabled

meson.build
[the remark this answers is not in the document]
nirbheek: Probably the other way around: a `desc` option that can be a free-form string. It is a convention in Meson that the first argument of a function specifies the name. For example: `library()` `dependency()` etc.

src/liblzma/meson.build
vtorri: so, on Windows:
* if get_id() returns 'msvc', i use the Windows linker option `/def:liblzma.def.in` (either link.exe, lld-link.exe or xilink.exe linkers)
* otherwise `-Wl,--output-def,liblzma.def.in`
is it ok for you ?

meson.build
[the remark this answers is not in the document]
vtorri: what I did for muon with cl.exe compiler: I just pass the required standard. So, no problem with c11. (there will be a message from the compiler if the standard is not correct)
I maybe should pass the standard if it is only c11 or c17 like meson. I'll ask the author

meson.build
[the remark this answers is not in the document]
vtorri: fixed in 070a65dc3c61a13447ec2de54c7f0fa7eea86087

meson.build
vtorri: fixed in 132cde1b031a0308a5408f7acca4ccaa9800dd36

meson.build
```diff
@@ -0,0 +1,343 @@
...
+# FIXME foreach i in array; set_variable(f'have_@i@', true) endforeach
+
+foreach f : get_option('encoders')
+    config_h.set10('HAVE_ENCODER_' + f.underscorify().to_upper(), true,
+        description: 'Define to 1 if ' + f + ' encoder is enabled.'
+    )
+    encoder_msg += f + ' '
+endforeach
+
+foreach f : get_option('encoders')
```
nirbheek, line 186:
```suggestion
foreach f : get_option('decoders')
```

vtorri: fixed

src/common/meson.build
vtorri: maybe the first list too can be improved (note that the variable cpu was used in the description of check line 230 of my original code) :

```diff --git a/src/common/meson.build b/src/common/meson.build
index d01853b1..ed3d3a75 100644
--- a/src/common/meson.build
+++ b/src/common/meson.build
@@ -182,15 +182,12 @@ if get_option('unaligned-access').auto()
         'x86',
         'x86_64',
     ]
-    foreach cpu : unaligned_access_cpus
-        if host_machine.cpu_family() == cpu
-            config_h.set10('TUKLIB_FAST_UNALIGNED_ACCESS', true,
-                description: 'Define to 1 if the system supports fast unaligned access to 16-bit, 32-bit, and 64-bit integers.'
-            )
-            fast_unaligned_access = true
-            break
-        endif
-    endforeach
+    if host_machine.cpu_family() in unaligned_access_cpus
+        config_h.set10('TUKLIB_FAST_UNALIGNED_ACCESS', true,
+            description: 'Define to 1 if the system supports fast unaligned access to 16-bit, 32-bit, and 64-bit integers.'
+        )
+        fast_unaligned_access = true
+    endif

     if not fast_unaligned_access
         unaligned_access_cpus = [
@@ -199,9 +196,8 @@ if get_option('unaligned-access').auto()
             'riscv32',
             'riscv64',
         ]
-        foreach cpu : unaligned_access_cpus
-            if host_machine.cpu_family() == cpu
-                ua_src = '''
+        if host_machine.cpu_family() in unaligned_access_cpus
+            ua_src = '''
 #if !defined(__ARM_FEATURE_UNALIGNED) \
                && !defined(__riscv_misaligned_fast) \
                && !defined(_MSC_VER)
@@ -209,15 +205,13 @@ compile error
 #endif
 int main(void) { return 0; }
 '''
-                if cc.compiles(ua_src, name: 'fast unaligned access for ' + cpu)
-                    config_h.set10('TUKLIB_FAST_UNALIGNED_ACCESS', true,
-                    description: 'Define to 1 if the system supports fast unaligned access to 16-bit, 32-bit, and 64-bit integers.'
-                    )
-                    fast_unaligned_access = true
-                    break
-                endif
+            if cc.compiles(ua_src, name: 'fast unaligned access for ' + host_machine.cpu_family())
+                config_h.set10('TUKLIB_FAST_UNALIGNED_ACCESS', true,
+                description: 'Define to 1 if the system supports fast unaligned access to 16-bit, 32-bit, and 64-bit integers.'
+                )
+                fast_unaligned_access = true
             endif
-        endforeach
+        endif
     endif
 elif get_option('unaligned-access').enabled()
     config_h.set10('TUKLIB_FAST_UNALIGNED_ACCESS', true,
```

src/liblzma/meson.build
Larhzu: The existing build systems create the DEF file only with MinGW-w64-based toolchains. No one has asked for it in MSVC builds so perhaps it's not important there. If one is building with MSVC, one might be building everything with it and thus not need DEF files for anything.

meson.build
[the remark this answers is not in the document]
Larhzu: The Meson docs give an impression that `c_std=gnu11` would fallback to C11 with MSVC but this might become deprecated and eventually an error. So one would need to list the possible standards.

I think requiring new enough Meson and using `c_std=gnu99,c11,c99` or `c_std=gnu11,gnu99,c11,c99` is fine for now. It might not be muon compatible right now but I would expect this kind of feature to get implemented at some point in muon too. Very probably the first versions of Meson support won't be as portable as the Autotools build is and it doesn't need to be. Progress happens one step at a time.

src/liblzma/meson.build
Larhzu: As discussed on IRC, it's fine to make LZMA1 and LZMA2 mandatory. Having a way to disable encoder support entirely is useful though.

meson.build
```diff
@@ -0,0 +1,343 @@
...
+            description: 'Define to 1 if the ' + f[0] + '() function is available.'
+        )
+        break
+    endif
+endforeach
+
+if cc.has_function('posix_advice', prefix : '#include <fcntl.h>')
+    config_h.set10('HAVE_POSIX_ADVICE', true,
+        description: 'Define to 1 if the posix_advice() function is available.'
+    )
```
Larhzu, line 157: It's `posix_fadvise`.

vtorri: fixed in 0ffc20d22329d3227c9773d5e1c347986ff85161

meson.build
```diff
@@ -0,0 +1,343 @@
+## SPDX-License-Identifier: 0BSD
+## Author: Vincent Torri
+
+project(
+    'xz-utils',
```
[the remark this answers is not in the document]
vtorri: ubuntu for example (https://packages.ubuntu.com/noble/xz-utils) does not respect the naming policy

fixed in aca1aa28d993d98a066e91940d46aba28079c4bf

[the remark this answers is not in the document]
Larhzu: Yes, Debian and thus distributions based on Debian use that name. Debian added 4.999.8beta to experimental three weeks before I committed PACKAGERS to xz.git. It's not a big deal at all. I know this project has a bit odd tarball name vs. pretty name so this happens easily.

meson.build
```diff
@@ -0,0 +1,350 @@
...
+config_h.set_quoted('PACKAGE', 'xz')
+config_h.set_quoted('PACKAGE_BUGREPORT', 'xz@tukaani.org')
+config_h.set_quoted('PACKAGE_NAME', meson.project_name())
+config_h.set_quoted('PACKAGE_URL', 'https://tukaani.org/xz/')
+
+# host
+
+config_h.set10('WORDS_BIGENDIAN', host_machine.endian() == 'big',
+    description: 'Define to 1 if the processor stores words with the most significant byte first.'
+)
```
Larhzu, line 41: `config_h.set10` always defines the macro to 1 or 0. The C code uses `#ifdef WORDS_BIGENDIAN`. Thus now it always thinks that it's on big endian. On little endian systems this results in liblzma that can decompress the files it creates (so certain tests will pass) but the files will be wrong. Testing with the included test files catches this (like the testing done by `test_files.sh`), thus this is a clear example why the source package needs test files.

`set10` is a good match when converting certain Autoconf macros like `AC_CHECK_DECLS` which always define the C macro to 1 or 0. Those feel a bit odd and we got rid of them in XZ Utils some time ago. I think Meson build of XZ Utils shouldn't use `set10` at all. Suggested code:

```
if host_machine.endian() == 'big'
    config_h.set('WORDS_BIGENDIAN', 1,
        description: 'Define to 1 if the processor stores words with the most significant byte first.')
endif
```

meson.build
```diff
@@ -0,0 +1,350 @@
...
+sys_windows = windows.contains(host_os)
+
+COND_W32 = sys_cygwin or sys_windows
+
+# binaries
+cc = meson.get_compiler('c')
+
+# try mimic AC_USE_SYSTEM_EXTENSIONS
+if host_machine.system() == 'linux'
+    add_project_arguments('-D_GNU_SOURCE', language: 'c')
```
Larhzu, line 60: Autoconf defines `_GNU_SOURCE` and the few other `AC_USE_SYSTEM_EXTENSIONS` macros unconditionally and it seems to be fine in practice. So I suggest simplifying this. Then GNU on something else than Linux (like HURD) will work without any extra effort too.

src/xzdec/meson.build
```diff
@@ -0,0 +1,89 @@
...
+        'xzdec.c'
+    ]
+
+    xzdec_cargs = [
+        '-DHAVE_CONFIG_H',
+        '-DHAVE_STDBOOL_H',
+        '-DHAVE_INTTYPES_H',
+        '-DHAVE_STDINT_H',
+        '-DTUKLIB_GETTEXT=0'
+    ]
```
Larhzu, line 17: I suggest puting `HAVE_STDBOOL_H`, `HAVE_INTTYPES_H`, and `HAVE_STDINT_H` into config.h in the top-level meson.build. The Meson build won't run on systems that lack them. After 5.8.0, XZ Utils will assume those to be present as well, making the macros unneeded.

Similarly, perhaps `HAVE_CONFIG_H` could be set at the top-level file with `add_project_arguments('-DHAVE_CONFIG_H')`.

`-DTUKLIB_GETTEXT=0` has to be kept here local to `xzdec` and `lzmadec`.

meson.build
```diff
@@ -0,0 +1,350 @@
...
+        '-Wsign-conversion',
+        '-Wfloat-conversion',
+        '-Wlogical-op',
+        '-Waggregate-return',
+        '-Wstrict-prototypes',
+        '-Wold-style-definition',
+        '-Wmissing-prototypes',
+        '-Wmissing-declarations',
+        '-Wredundant-decls',
+#
```
Larhzu, line 98: This `#` should be indented. Now it reminds me of a `.` at the beginning of a line. ;-)

src/common/meson.build
```diff
@@ -0,0 +1,384 @@
...
+    config_h.set10('HAVE___BUILTIN_ASSUME_ALIGNED', true,
+        description: 'Define to 1 if the GNU C extension __builtin_assume_aligned is supported.'
+                    )
+endif
+
+# tuklib_mbstr.m4
+
+### FIXME : mbrtowc is supported in C99 spec and on Windows
+if cc.has_header('wchar.h')
+    config_h.set10('HAVE_WCHAR_H', true,
```
Larhzu, line 233: Nothing in the package needs `HAVE_WCHAR_H`.

src/common/meson.build
```diff
@@ -0,0 +1,384 @@
...
+
+if get_option('unsafe-type-punning')
+    config_h.set10('HAVE___BUILTIN_ASSUME_ALIGNED', true,
+        description: 'Define to 1 if the GNU C extension __builtin_assume_aligned is supported.'
+                    )
+endif
+
+# tuklib_mbstr.m4
+
+### FIXME : mbrtowc is supported in C99 spec and on Windows
```
Larhzu, line 231: This is true. When I added the check, I felt it was the simplest way to ensure that the code will keep working on odd platforms. [Gnulib's mbrtowc](https://www.gnu.org/software/gnulib/manual/html_node/mbrtowc.html) lists _old versions_ of Minix 3 and HP-UX and also IRIX as lacking `mbrtowc`. I don't think those matter in Meson build at all. The latest MINIX 3 does have `mbrtowc` and so does HP-UX. IRIX is too old.

So it's possible that, among the supported or potentially-supported platforms, only DJGPP lacks `mbrtowc`. If so, it could be handled with a special `#ifdef` for that in the C code. But it's also not much work to check it in the build system.

I currently believe it's fine if the Meson build unconditionally defines `HAVE_MBRTOWC`. It might be cleaned up later along with `HAVE_STDINT_H` and such.

meson.build
```diff
@@ -0,0 +1,350 @@
...
+static void my_constructor_func(void) { return; }
+int main() { return 0; }
+'''
+if cc.compiles(ctor_fct_src, name : 'constructor attribute for functions')
+    config_h.set10('HAVE_FUNC_ATTRIBUTE_CONSTRUCTOR', true,
+        description: 'Define to 1 if __attribute__((__constructor__)) is supported for functions.'
+    )
+endif
+
+cc.has_type('uint32_t', prefix: '#include <stdint.h>')
```
Larhzu, line 135: This looks like a no-op as nothing checks the return value and `required: true` isn't set. The Meson build can assume that `<stdin.h>` and the standard types there are available, so this line can be removed.

meson.build
Larhzu: Correction: Due to the use of the unmodified visibility.m4 from Gnulib, `HAVE_VISIBILITY` has to be always defined and thus use `set10`.

src/common/meson.build
```diff
@@ -0,0 +1,384 @@
...
+    )
+
+    if cc.has_type('mbstate_t', prefix : '#include <wchar.h>') and cc.has_function('mbrtowc', prefix : '#include <wchar.h>')
+        config_h.set10('HAVE_MBRTOWC', true,
+            description: 'Define to 1 if the mbrtowc() function is available.'
+        )
+    endif
+
+    if cc.has_function('wcwidth',
+        args : '-D_XOPEN_SOURCE',
```
Larhzu, line 244: It's incorrect to add `-D_XOPEN_SOURCE` into a single test. The actual C code won't use that `#define`, and thus the test and the C code wouldn't be in sync. If you added this because the wcwidth(3) man page in the man-pages package lists it, note that the same line on the man page tells to look at feature_test_macros(7) which tells that `_GNU_SOURCE` already covers this.

The codebase relies on some platform-specific extensions when they are available, thus it's not practical to build it with something like `-D_XOPEN_SOURCE=600`. Instead, all system-specific extensions are hoped to be available which means using, for example, `#define _GNU_SOURCE 1`. Also, some of the now-POSIX features were available on some systems before they were standardized in POSIX, so this approach works better on such systems.

Thus, it's essential that `_GNU_SOURCE` and such things are added to project-wide defines before running any checks for available functions or other features, and the checks then don't use any function-specific extra args unless the C code uses such extras too.

meson.build
eli-schwartz: I have always felt the opposite regarding weirdness. If you use ifdef, you can't tell whether files are supposed to not have it defined or whether the build system forgot to check it / config.h wasn't included in that file. If it is always guaranteed to be defined, and you check which state it has with `#if`, then you can take advantage of -Wundef.

Still, the port must accommodate the code that exists, not the code that contributors would prefer...

Larhzu: I agree. I meant weird in context of Autoconf where almost all other checks either define or not define a macro instead of always defining to 1 or 0. I don't like mixing both styles because due to `-Wundef` one cannot use `#if` to check those macros that aren't always defined. When both styles are present, one has to be careful to know which macros to check with `#ifdef` and which with `#if`. Nowadays `HAVE_VISIBILITY` is such a speciality in XZ Utils.

Converting the codebase to only use macros that are always defined is possible but it's quite a bit of work too and there is a risk of errors. It would also affect build systems that are maintained outside the project. Maybe it could be done some day, who knows, but it's not a high priority for now.

src/common/meson.build
```diff
@@ -0,0 +1,384 @@
...
+                )
+            endif
+        endforeach
+    endif
+endif
+
+fast_unaligned_access = false
+if get_option('unaligned-access').auto()
+    unaligned_access_cpus = [
+        'ppc',
```
Larhzu, line 180: The Autoconf and CMake builds exclude 32-bit PowerPC from fast unaligned access as I'm not sure how it behaves, especially with 64-bit values. I suppose it's an uncommon target anyway so better play it safe.

meson.build
```diff
@@ -0,0 +1,350 @@
...
+        '-Wshift-sign-overflow',
+        '-Wstring-conversion',
+    ]
+
+    lzma_c_args += cc.get_supported_arguments(cflags_try)
+endif
+
+config_h.set10('HAVE_VISIBILITY', cc.has_function_attribute('visibility'),
+    description: 'Define to 1 if the compiler supports simple visibility declarations.'
+)
```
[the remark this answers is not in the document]
eli-schwartz: Meson internally handles -Werror for exactly that reason:
.https://github.com/mesonbuild/meson/blob/7d28ff29396f9d7043204de8ddc52226b9903811/mesonbuild/compilers/mixins/clike.py#L1309-L1324

Most projects I have seen, actually check for `__GNUC__` instead of using visibility.m4 even with autotools. This is kind of necessary anyway since you need preprocessor logic to figure out whether Microsoft dllexport/dllimport is viable, and visibility.m4 is ultimately not all that helpful overall.


Certain versions of... SunPro, if I recall correctly, are the only commonly known compiler that handles visibility but isn't either `__GNUC__` or `MSC_VER`, and they use a third syntax instead?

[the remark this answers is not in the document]
Larhzu: > Meson internally handles -Werror for exactly that reason:

Great! Documenting it could be nice to give more confidence. Or perhaps it's that coming from other build systems one is constantly suspecting that corner cases aren't handled well enough, sorry.

> Most projects I have seen, actually check for `__GNUC__` instead of using visibility.m4 even with autotools.

liblzma checks in this order in the C code:

1. `#if defined(_WIN32) || defined(__CYGWIN__)` then use `__declspec(dllexport)`.

2. `#if HAVE_VISIBILITY` to use GNU C attributes. Checking for `__GNUC__` wouldn't work as it can be defined on platforms that don't support the visibility attributes; one has to know if the attribute is actually supported.

3. Otherwise assume no visibility support.

> Certain versions of... SunPro, if I recall correctly, are the only commonly known compiler that handles visibility but isn't either `__GNUC__` or `MSC_VER`, and they use a third syntax instead?

My note in b0d1422b6037bfea6f6723683bd82a8e6d77026c says that SunPro (or whatever its name is this year) supports the GNU C attribute since 2007. At least modern releases support the `-fvisibility` option too. I didn't research when the command line option was added (it would make sense to add it when the attribute was but I don't know if that happened). The original option is `-xldscope` but I guess there's no point to add support for it when `-fvisibility` works too.

meson_options.txt
```diff
@@ -0,0 +1,188 @@
...
+    'unaligned-access',
+    type: 'feature',
+    value: 'auto',
+    description: '*fast* unaligned memory access with 16-bit, 32-bit, and 64-bit integers. Enabled by default on  x86, x86-64, 32/64-bit big endian PowerPC, 64-bit little endian PowerPC and some ARM, ARM64, and RISC-V systems.'
+)
+
+option('unsafe-type-punning',
+    type : 'boolean',
+    value : false,
+    description : 'use ARM64 CRC32 instructions'
```
Larhzu, line 175: A copy-paste error

meson_options.txt
```diff
@@ -0,0 +1,188 @@
+## SPDX-License-Identifier: 0BSD
+## Author: Vincent Torri
```
Larhzu, line 2: Makefile.am files use `##` because Automake omits such lines from the produced Makefile.in files; lines with a single `#` get into Makefile.in. So there is no need to use `##` in meson.build unless it was a stylistic preference.

meson.build
Larhzu: Autoconf's `AC_DEFINE` always puts something into config.h, either `#define FOO 1` or `/* #undef FOO */`. This way it's easy to see both what was and what wasn't defined.

Meson cannot do that. Thus in Meson context the 1/0 method has an extra benefit on top of what was already mentioned.

meson.build
```diff
@@ -0,0 +1,350 @@
...
+ctor_fct_src = '''
+__attribute__((__constructor__))
+static void my_constructor_func(void) { return; }
+int main() { return 0; }
+'''
+if cc.compiles(ctor_fct_src, name : 'constructor attribute for functions')
+    config_h.set10('HAVE_FUNC_ATTRIBUTE_CONSTRUCTOR', true,
+        description: 'Define to 1 if __attribute__((__constructor__)) is supported for functions.'
+    )
+endif
```
Larhzu, line 133: Simpler:

```
if cc.has_function_attribute('constructor')
    config_h.set('HAVE_FUNC_ATTRIBUTE_CONSTRUCTOR', 1,
        description: 'Define to 1 if __attribute__((__constructor__)) is supported for functions.')
endif
```

meson.build
eli-schwartz: With set() instead of set10, it expects either true or false and true will `#define FOO` but not set it to 1 -- false will #undef it.

Larhzu: Ah, thanks! Now I wonder which I prefer. The explicit `#undef` has a clearer meaning than Autoconf's commented version, but Autoconf's version is visually easier to spot in the list. The explicit one is trivial to filter to something else _for viewing purposes_ so it's not a problem. :-)

I think I've been careful to never assume that a macro has a value excluding the special cases. And an empty value in `#if` would usually be an error. So perhaps the `set` method with true or false is nicer. It avoids `if`...`endif` clutter in `meson.build` files.

meson.build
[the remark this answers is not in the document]
Larhzu: This

```
__attribute__((dllexport)) int foo(int x) { return x; }
```

results in a warning with GCC 14.1.1 on x86-64 GNU/Linux:

```
warning: ‘dllexport’ attribute directive ignored [-Wattributes]
    1 | __attribute__((dllexport)) int foo(int x) { return x; }
      | ^~~~~~~~~~~~~
```

But `cc.has_function_attribute('dllexport')` is still true. So `-Werror` isn't used or Meson shortcircuits the attribute check somehow. Thus I fear it might do the same with the visibility attribute as well.

meson_options.txt
vtorri: i'll change that if you want. I just copy/pasted this from muon and there was 2 #

meson.build
vtorri: fixed in 96e2d962ba531e7e2d4e953ed0bef12a8dbe7f45

src/common/meson.build
vtorri: fixed in f51a11f845a064f8410dd1cb3411dfd0344d0dad

src/common/meson.build
vtorri: done in f51a11f845a064f8410dd1cb3411dfd0344d0dad

src/xzdec/meson.build
vtorri: I move them at the top level meson.build for the whole project ( in f51a11f845a064f8410dd1cb3411dfd0344d0dad)

meson.build
vtorri: I see no '.' at the beginning of a line in toplevel meson.build. fix # indent in be2a23a758e387583b8edf95443b33e18e6bc68f

meson_options.txt
vtorri: done in 8a7b9b11de71c9b0511a4a70d44fe18c1dd7e6fd

meson.build
vtorri: done in a83a638d7a129a69f54e5c9a0b86eee96dfaaa56

meson.build
vtorri: so is

```
config_h.set('WORDS_BIGENDIAN', host_machine.endian() == 'big',
    description: 'Define to 1 if the processor stores words with the most significant byte first.'
)
```

(which generates `#undef WORDS_BIGENDIAN` on little endian) good for you ?

src/common/meson.build
thesamesam: I, of course, agree with you on all of the above - in addition to it being cleaner to be consistent, but it's _also_ an issue for C++ with ODR too. And sometimes projects have to be built as C++ even if they're written in C.

Larhzu:
> (which generates `#undef WORDS_BIGENDIAN` on little endian) good for
> you ?

Yes. Thanks!

Larhzu:
> I see no '.' at the beginning of a line in toplevel meson.build.

True and I didn't mean to imply that. In case you missed my point, see
the commit f9cf4c05edd14dedfe63833f8ccbe41b55823b00.

vtorri:
> (which generates `#undef WORDS_BIGENDIAN` on little endian) good for you ?
> Yes. Thanks!

done

meson.build
vtorri: done in fb6b458e9ed311ed1728de3b4c4617fd8d5861d3

meson.build
vtorri: done in fb6b458e9ed311ed1728de3b4c4617fd8d5861d3

meson_options.txt
vtorri: fixed in efcb02e6fa110cbadac1b1a84fd2ba3674983142

PR #124, opened 2024-06-05
fix typos

RainRat:


Larhzu:
> M doc/xz-file-format.txt (2)

I only queued this locally because I don't want to make a new
xz-file-format.txt release solely for one typo fix.

Thanks!

Issue #129, opened 2024-07-02
CMake configure error in case '-DENABLE_NLS=ON' is given and cmake version is to old (e.g. 3.18.4)


pseiderer:
CMake configure error incase '-DENABLE_NLS=ON' is given and cmake version is to old (e.g. 3.18.4):

      CMake Error at CMakeLists.txt:1675 (add_executable):
        Target "xz" links to target "Intl::Intl" but the target was not found.
        Perhaps a find_package() call is missing for an IMPORTED target, or an
        ALIAS target is missing?
    
      CMake Error at CMakeLists.txt:1619 (add_executable):
        Target "lzmainfo" links to target "Intl::Intl" but the target was not
        found.  Perhaps a find_package() call is missing for an IMPORTED target, or
        an ALIAS target is missing?

Larhzu:
I suppose it's not the best if user explicitly enables NLS and that gets silently ignored. Here the current code is buggy with new CMake too. I try to make an alternative fix proposal.

Larhzu:
Please checkout a9950d573cd5f6ae6940992019fd38d564bb014c in the `cmake_nls_fixes` branch. It changes the default behavior a little too. The default is still a bit weird as it depends on whether `Intl` is found or not. But perhaps it's better this way still. In any case, it should fix bugs when user sets `XZ_NLS=ON` but when it cannot actually work due to too old CMake or missing `Intl`.

I would like to hear your opinion if this is OK/better/worse than your suggestion.

I'm also pondering if support for pre-generated `.gmo` files should be dropped. It's very little code and simple code too, but `.gmo` files are binary files that get installed. If people can install CMake, I suppose they can install gettext tools too. Shipping pre-generated `.gmo` files feels like Autotools-specific thing.

Larhzu:
A few days ago I posted [in the other issue](https://github.com/tukaani-project/xz/issues/68#issuecomment-2188756570) if it was OK to require CMake 3.20. I suppose there was a reason why you tested with an older version, so it would be valuable to hear how inconvenient it would be to you if 3.20 was required. Thanks!

pseiderer:
Yep, branch cmake_nls_fixes works for me (and a warning/error is definitely better)...

I believe dropping pre-generated .gmo files and gettext support would be o.k. for the Buildroot target package build (my use case)...

Require CMake 3.20 would be a bit problematic as in Buildroot there is not (yet) support for host packages to require a minimum CMake version (for the target packages support is present)...

[...]

PR #130, opened 2024-07-02
cmake: overwrite user given NLS decision in case legacy cmake is detected

pseiderer:
Overwriteuser given NLS decision in case legacy cmake (e.g. 3.18.4) is detected.

Fixes:

  CMake Error at CMakeLists.txt:1675 (add_executable):
    Target "xz" links to target "Intl::Intl" but the target was not found.
    Perhaps a find_package() call is missing for an IMPORTED target, or an
    ALIAS target is missing?

  CMake Error at CMakeLists.txt:1619 (add_executable):
    Target "lzmainfo" links to target "Intl::Intl" but the target was not
    found.  Perhaps a find_package() call is missing for an IMPORTED target, or
    an ALIAS target is missing?

Closes: https://github.com/tukaani-project/xz/issues/129

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -360,6 +360,9 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.20")
         # The *installed* name of the translation files is "xz.mo".
         set(TRANSLATION_DOMAIN "xz")
     endif()
+else()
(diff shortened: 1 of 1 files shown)


thesamesam:
@Larhzu I don't want to say "I told you so..." :D

PR #134, opened 2024-08-06
Fix: XZ_TOOL_LZMADEC

wzhengsen:
When the XZ_TOOL_LZMADEC option is turned off, lzmadec should not be used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1868,14 +1868,16 @@ if(HAVE_DECODERS)
                         COMPONENT "${XZDEC}_Runtime")
     endforeach()
 
-    # This is the only build-time difference with lzmadec.
-    target_compile_definitions(lzmadec PRIVATE "LZMADEC")
+    if(XZ_TOOL_LZMADEC)
+        # This is the only build-time difference with lzmadec.
+        target_compile_definitions(lzmadec PRIVATE "LZMADEC")
 
-    if(UNIX)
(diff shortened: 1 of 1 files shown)


Larhzu:
Thanks! Clearly I hadn't tested that option at all.

I created a branch `cmake_lzmadec`. It differs in how the xzdec/lzmadec man page installation is handled when the tools are disabled. Your version installed the xzdec man page and the `lzmadec.1` symlink when `XZ_TOOL_XZDEC=OFF`. My version is kind of the opposite as it doesn't install any man page in that situation. So it's not perfect either. But I think it's a good enough compromise while keeping things still simple.

* * *

How would you like to be credited, if at all, in THANKS and commit log? The way I did it in `cmake_lzmadec` is fine for me. However, I'm asking because in English context people often seem to transliterate their names to Latin script. That way people who don't know, for example, Chinese, Cyrillic, or Greek scripts can recognize the names easier and have a chance to at least mispronounce them (instead of not being able to pronounce them at all).

I apologize if this is a sensitive question. I just wish to understand the topic better. Thanks!

wzhengsen:
@Larhzu Thank you! If you need to use an English name, you can use my “wzhengsen” (which is just a phonetic translation of my Chinese name and what I usually use)

Larhzu:
Thank you! I added it in parenthesis. :-)

PR #137, opened 2024-09-02
Add support for Large Memory  Model on AIX 32-bit

vchaves123:
32-bit executables on AIX need to increase the address space to allow more than 256MB RAM.
It would be nice if this was available out of box.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -57,6 +57,7 @@ AC_SUBST([LN_EXEEXT])
 echo
 echo "Configure options:"
 AM_CFLAGS=
+AM_LDFLAGS=
 
 
 #############
@@ -868,6 +869,13 @@ AC_CHECK_MEMBERS([
 AC_SYS_LARGEFILE
 AC_C_BIGENDIAN
 
+case $host_os in
+	aix*)
+		if test "x$gl_cv_host_cpu_c_abi_32bit" = xyes ; then
+			AM_LDFLAGS="$AM_LDFLAGS -Wl,-bmaxdata:0x80000000"
+		fi
+esac
+
 # __attribute__((__constructor__)) can be used for one-time initializations.
 # Use -Werror because some compilers accept unknown attributes and just
 # give a warning.
@@ -1323,6 +1331,7 @@ AM_CONDITIONAL([COND_GNULIB], test -n "$LIBOBJS")
 
 # Add default AM_CFLAGS.
 AC_SUBST([AM_CFLAGS])
+AC_SUBST([AM_LDFLAGS])
 
 # This is needed for src/scripts.
 xz=`echo xz | sed "$program_transform_name"`


Larhzu:
Do all linkers on 32-bit AIX support the -bmaxdata option? For example, I don't find it from GNU ld documentation. Maybe it needs to be handled conditionally.

gl_cv_host_cpu_c_abi_32bit comes from host-cpu-c-abi.m4. That file gets inserted to the xz package with gettext >= 0.20 which has picked it from Gnulib. If that is used, the macro gl_HOST_CPU_C_ABI_32BIT should be called explicitly to set gl_cv_host_cpu_c_abi_32bit, and the .m4 file likely should be committed to the repository as well, and then kept up to date. Specifically, it's not OK to rely on gettext's macros calling it as a future version of gettext isn't guaranteed to do so.

Maybe checking gl_cv_host_cpu_c_abi_32bit a good way to do this. However, I wonder if checking host_cpu would work too, or perhaps checking if ac_cv_sizeof_size_t equals 4. It is set by AC_CHECK_SIZEOF([size_t]) which is already in configure.ac.

CMake side should be changed to match but that should be easy enough. So let's worry about that once we know what to do.

I assume you have seen the AIX-specific notes in INSTALL. It only mentions the usage of CC=xlc_r. m4/ax_pthread.m4 attempts to autodetect it but it's not used in configure.ac because back in the day it seemed troublesome because CC is being modified by AC_PROG_CC_C99 and CC might contain also arguments, not just a command name. I don't know if that xlc_r note is still relevant; I haven't tested on AIX.

While I get that it can be nice if the address space setting was done by default, I also consider the alternative of adding a note to INSTALL and requiring people to set it manually. This is the first time I hear about this, so perhaps it's fairly common knowledge to set such LDFLAGS. I mean, I assume several other packages require adding the same LDFLAGS so doing so it might be a kind of standard thing to do. I'm just guessing now, trying to understand the big picture.

Thanks!

vchaves123:
Hello Lasse,
my setup consists of the GCC 10 compiler and the native IBM linker. I wouldn't be able to say if the GNU linker supports the maxdata flag.

Yes, I saw the notes about AIX in the INSTALL file, but they apply to those who use the paid IBM xlc compiler and need to use the xlc_r binary for programs that make use of the pthreads library.

My knowledge of autoconf is limited, but it seems to me that using sizeof is perfectly viable.
In fact, the most correct pseudo logic should be: 

if AIX and  sizeof=4 and GCC and IBM ld 
then 
 set LDFLAGS

Again, it would be great to see the old AIX supported out of the box, but the best course of action is up to your discretion.

Sincerely,
Vicente.

Larhzu:
I'm curious, how old AIX and hardware is it that 64-bit binaries aren't possible (or practical)?

[PostgreSQL's platform-specific notes](https://www.postgresql.org/docs/current/installation-platform-notes.html) mention address space configuration on 32-bit AIX. Python mentions `maxdata` in [README.AIX](https://github.com/python/cpython/blob/main/Misc/README.AIX). It seems that these packages don't set the option by default.

Maybe setting this option by default could make sense but it's also a topic that people building on AIX will encounter with other packages too. Thus, it's somewhat likely that people are (or will be) aware of this special case on 32-bit AIX. Special handling in one package (xz) might have unwanted effects when experienced packagers are used to setting the option themselves, although perhaps this isn't a very likely problem.

This PR suggests using `-Wl,-bmaxdata:0x80000000`. Reading AIX 7.3 docs, that seems to allow 2 GiB of address space using _large address-space model_. But `-Wl,-bmaxdata:0xD0000000/dsa` would allow 3.25 GiB using _**very** large address-space model_. Shouldn't the latter be recommended? Or is it not available in old AIX versions? AIX 7.1 docs mention it still and 7.1 support ended in 2023.

The docs say that values higher than 0xAFFFFFFF have a downside of not sharing memory of shared libraries. So perhaps 0xD0000000 isn't ideal here. I don't have much clue because I haven't used AIX. 0xAFFFFFFF would still be 2.75 GiB.

It shouldn't be too hard to autodetect if the option is supported by just checking if linking works with those `LDFLAGS`. One would need to verify that user-defined `LDFLAGS` would always override the default value. But right now I feel it's enough to give a clue about this in INSTALL. How about:

> If building a 32-bit executable, the address space available for xz might be limited to 256 MiB. To allow 2.75 GiB address space, pass LDFLAGS='-Wl,-bmaxdata:0xAFFFFFFF/dsa' as an argument to configure. For more information, see "Large program support" in the AIX documentation. For example, AIX 7.3 documentation is available at the following address (the link was checked on 2024-09-06):
>
> https://www.ibm.com/docs/en/aix/7.3?topic=concepts-large-program-support

Improvement ideas for the text are welcome. I suppose the main question is, what value to suggest by default. If you have a 400-megabyte file, compressing it with `xz -9 -T2` should work with 0xAFFFFFFF but not with 0x80000000. Can you test this?

dcondoris (approved):

PR #140, opened 2024-09-08
Translations: Fix improper expressions in Chinese (simplified) translation.

ChenPi12:


Larhzu:
The xz translations are handled by the Translation Project. That is, I take updated PO files from there, and any fixes must go via the Translation Project too. The translators *don't* pick updated PO files from the xz source tree, so if I merged your fixes directly, they would likely get lost later.

Please report issues in the translated strings to the language team and to the last translator as well. If there is a problem, let me know. Thank you!

PR #143, opened 2024-09-16
xzdec: Remove unused short option

stoeckmann:
The "M:" entry must be a leftover from xz tool.
It's not available in xzdec.


Larhzu:
Well spotted! Thanks! It's a left-over from 792331bdee706aa852a78b171040ebf814c6f3ae.

PR #144, opened 2024-09-16
lzmainfo: Avoid integer overflow

stoeckmann:
The MB output can overflow with huge numbers. Most likely these are invalid lzma files anyway, but let's avoid garbage output.

Proof of Concept:

1. Create an lzma file with huge values
```
base64 -d <<< AP/////+/////////w== > poc.lzma
```

2. Run lzmainfo
```
lzmainfo poc.lzma
```

You can see this output:
```
poc.lzma
Uncompressed size:             0 MB (18446744073709551614 bytes)
Dictionary size:               0 MB (2^31 bytes)
Literal context bits (lc):     0
Literal pos bits (lp):         0
Number of pos bits (pb):       0
```

Clearly these MB values are wrong.


Larhzu:
Thanks! The relevant code goes back to early LZMA Utils from 2005, so the bug is over 19 years old. This has to be the oldest bug that has been reported.

The lowest bits get ignored so could it be simplified? `(bytes / 1024 + 512) / 1024`

stoeckmann:
> The lowest bits get ignored so could it be simplified? `(bytes / 1024 + 512) / 1024`

That makes the calculation much easier! At least if my memories of school's math classes are still correct. :)

Larhzu:
Thanks!

PR #155, opened 2024-12-08
liblzma: Fix incorrect macro name in a comment

dextercd:
Very minor thing but this tripped me up. The comment references `lzma_ext_size_set` but this doesn't exist. It's supposed to say `lzma_set_ext_size`, with set at the start not the end.


Larhzu:
Thanks! Misleading docs are annoying so it's good to fix this kind of things. :-)

PR #166, opened 2025-02-16
Build: Add support for `gl_REPLACE_GETOPT_ALWAYS` in getopt.m4

lzcunt:
Gnulib supports forcing the replacement of getopt & getopt_long, but that was removed from the in-tree simplified getopt.m4. This is useful for building for targets with broken getopt_long implementations.

### DIFF
--- a/m4/getopt.m4
+++ b/m4/getopt.m4
@@ -16,6 +16,10 @@ AC_DEFUN([gl_FUNC_GETOPT_GNU],
 [
   AC_REQUIRE([gl_GETOPT_CHECK_HEADERS])
 
+  dnl Other modules can request the gnulib implementation of the getopt
+  dnl functions unconditionally, by defining gl_REPLACE_GETOPT_ALWAYS.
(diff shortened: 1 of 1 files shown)


lzcunt:
Thanks for the feedback! I'm not good with autotools so I simply copied what upstream gnulib was doing, but in hindsight this seems like a better solution.

Thanks for maintaining this awesome project, much love 💗

Larhzu:
This was a good improvement for testing purposes too. Thanks to this, I found the out-of-tree build issue. If you like, I can add your real name (not a nick or screen name) to THANKS. Every small fix counts. :-)

lzcunt:
Thanks the offer, but at this time I don't enjoy putting my name over stuff. I don't really keep it private - my git email has my legal name - but I don't  put it in text for several reasons

Larhzu:
OK. :-) Several people have a similar preference to not be named, which of course is fine. Thanks!

PR #168, opened 2025-03-10
tuklib_physmem: Silence -Wsign-conversion on AIX.

collinfunk:
On AIX:

```
$ ./configure && make
[...]
../common/tuklib_physmem.c:151:8: warning: conversion to 'uint64_t' {aka 'long long unsigned int'} from 'long long int' may change the sign of the result [-Wsign-conversion]
  151 |  ret = _system_configuration.physmem;
      |        ^~~~~~~~~~~~~~~~~~~~~
mv -f .deps/liblzma_la-tuklib_physmem.Tpo .deps/liblzma_la-tuklib_physmem.Plo
```

This is harmless, but since the warning is enabled by ./configure if available I figured you would want it fixed.

### DIFF
--- a/src/common/tuklib_physmem.c
+++ b/src/common/tuklib_physmem.c
@@ -148,7 +148,7 @@ tuklib_physmem(void)
 			ret += entries[i].end - entries[i].start + 1;
 
 #elif defined(TUKLIB_PHYSMEM_AIX)
-	ret = _system_configuration.physmem;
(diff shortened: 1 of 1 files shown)


Larhzu:
Thanks! I like to fix these because in some cases -Wsign-conversion catches real mistakes.

I also committed a note about a linker flag on 32-bit AIX:
80e48836024ec2d7cbd557575be6da3d1f055cba

PR #173, opened 2025-04-25
liblzma: Add LoongArch BCJ filter

lrzlin:
This patch adds LoongArch BCJ filter support, it has been tested on real Loongson hardware & QEMU.

It's mostly refer the arm64's filter, converts BL/PCALAU12I only, other BCJ instructions was also tested but they do not have positive effet on redundancy. The PCALAU12I's range is also limited to +/-512MB to reduce false positives.

However, when testing BCJ filter on binaries with debug info, the compress ratio significant worse than use LZMA2 only, It's majority due to many false positives of BL instructions, so I've add a if-statement in BL part that only if the previous instructions of BL is in a range of instructions we selected, we'll convert it's immediate to absolute address, or we'll just ignores it. These instructions were derived from analyzing LoongArch architecture's libraries and executables.

This optimize could largely improves compress ratio when dealing with this kinds of libraries or executables. But this optimize relies on global varibales, I don't sure whether it's reliable in multithread environment. Currently, an one-hour test on Loongson 3C6000(16C/32T) shows no data corruption. It's still a very naive attempt, so it would be much appreciated if you could provide some feedback.

Also, I used 0x0C as this filter's ID, but it's just a temporary path to test the filter.

Related: https://github.com/loongson-community/discussions/issues/78

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -935,6 +935,7 @@ set(SIMPLE_FILTERS
     ia64
     sparc
     riscv
+    loongarch
 )
 
 # The SUPPORTED_FILTERS are shared between Encoders and Decoders
--- a/build-aux/ci_build.bash
+++ b/build-aux/ci_build.bash
@@ -197,7 +197,7 @@ then
 
 		cd "$DEST_DIR"
 
-		add_to_filter_list "$BCJ" ",x86,powerpc,ia64,arm,armthumb,arm64,sparc,riscv"
+		add_to_filter_list "$BCJ" ",x86,powerpc,ia64,arm,armthumb,arm64,sparc,riscv,loongarch"
 		add_to_filter_list "$DELTA" ",delta"
 
 		add_extra_option "$ENCODERS" "--enable-encoders=$FILTER_LIST" "--disable-encoders"
@@ -227,7 +227,7 @@ then
 	cmake)
 		cd "$DEST_DIR"
 
-		add_to_filter_list "$BCJ" ";x86;powerpc;ia64;arm;armthumb;arm64;sparc;riscv"
+		add_to_filter_list "$BCJ" ";x86;powerpc;ia64;arm;armthumb;arm64;sparc;riscv;loongarch"
 		add_to_filter_list "$DELTA" ";delta"
 
 		add_extra_option "$THREADS" "-DXZ_THREADS=yes" "-DXZ_THREADS=no"
--- a/configure.ac
+++ b/configure.ac
@@ -78,8 +78,8 @@ fi
 # Filters #
 ###########
 
-m4_define([SUPPORTED_FILTERS], [lzma1,lzma2,delta,x86,powerpc,ia64,arm,armthumb,arm64,sparc,riscv])dnl
-m4_define([SIMPLE_FILTERS], [x86,powerpc,ia64,arm,armthumb,arm64,sparc,riscv])
+m4_define([SUPPORTED_FILTERS], [lzma1,lzma2,delta,x86,powerpc,ia64,arm,armthumb,arm64,sparc,riscv,loongarch])dnl
+m4_define([SIMPLE_FILTERS], [x86,powerpc,ia64,arm,armthumb,arm64,sparc,riscv,loongarch])
 m4_define([LZ_FILTERS], [lzma1,lzma2])
 
 m4_foreach([NAME], [SUPPORTED_FILTERS],
--- a/src/liblzma/api/lzma/bcj.h
+++ b/src/liblzma/api/lzma/bcj.h
@@ -57,6 +57,10 @@
  */
 #define LZMA_FILTER_RISCV       LZMA_VLI_C(0x0B)
 
+/**
+ * \brief       Filter for LoongArch binaries
+ */
+#define LZMA_FILTER_LOONGARCH   LZMA_VLI_C(0x47BEC794C61203)
 
 /**
  * \brief       Options for BCJ filters
@@ -193,3 +197,34 @@ extern LZMA_API(size_t) lzma_bcj_x86_encode(
  */
 extern LZMA_API(size_t) lzma_bcj_x86_decode(
 		uint32_t start_offset, uint8_t *buf, size_t size) lzma_nothrow;
+
+
+/**
+ * \brief       Raw LoongArch BCJ encoder
+ *
+ * This is for special use cases only.
+ *
+ * \param       start_offset  The lowest 32 bits of the offset in the
+ *                            executable being filtered. For the LoongArch
+ *                            filter, this must be a multiple of four.
+ * \param       buf           Buffer to be filtered in place
+ * \param       size          Size of the buffer
+ *
+ * \return      Number of bytes that were processed in `buf`. This is at most
+ *              `size`. With the LoongArch filter, the return value is always
+ *              a multiple of 4, and at most 3 bytes are left unfiltered.
+ *
+ * \since       5.8.1alpha
+ */
+extern LZMA_API(size_t) lzma_bcj_loongarch_encode(
+	uint32_t start_offset, uint8_t *buf, size_t size) lzma_nothrow;
+
+/**
+* \brief       Raw LoongArch BCJ decoder
+*
+* See lzma_bcj_loongarch_encode().
+*
+* \since       5.8.1alpha
+*/
+extern LZMA_API(size_t) lzma_bcj_loongarch_decode(
+	uint32_t start_offset, uint8_t *buf, size_t size) lzma_nothrow;
--- a/src/liblzma/common/filter_common.c
+++ b/src/liblzma/common/filter_common.c
@@ -130,6 +130,15 @@ static const struct {
 		.changes_size = false,
 	},
 #endif
+#if defined(HAVE_ENCODER_LOONGARCH) || defined(HAVE_DECODER_LOONGARCH)
+	{
+		.id = LZMA_FILTER_LOONGARCH,
+		.options_size = sizeof(lzma_options_bcj),
+		.non_last_ok = true,
+		.last_ok = false,
+		.changes_size = false,
+	},
+#endif
 #if defined(HAVE_ENCODER_DELTA) || defined(HAVE_DECODER_DELTA)
 	{
 		.id = LZMA_FILTER_DELTA,
--- a/src/liblzma/common/filter_decoder.c
+++ b/src/liblzma/common/filter_decoder.c
@@ -128,6 +128,14 @@ static const lzma_filter_decoder decoders[] = {
 		.props_decode = &lzma_simple_props_decode,
 	},
 #endif
+#ifdef HAVE_DECODER_LOONGARCH
+	{
+		.id = LZMA_FILTER_LOONGARCH,
+		.init = &lzma_simple_loongarch_decoder_init,
+		.memusage = NULL,
+		.props_decode = &lzma_simple_props_decode,
+	},
+#endif
 #ifdef HAVE_DECODER_DELTA
 	{
 		.id = LZMA_FILTER_DELTA,
--- a/src/liblzma/common/filter_encoder.c
+++ b/src/liblzma/common/filter_encoder.c
@@ -168,6 +168,16 @@ static const lzma_filter_encoder encoders[] = {
 		.props_encode = &lzma_simple_props_encode,
 	},
 #endif
+#ifdef HAVE_ENCODER_LOONGARCH
+	{
+		.id = LZMA_FILTER_LOONGARCH,
+		.init = &lzma_simple_loongarch_encoder_init,
+		.memusage = NULL,
+		.block_size = NULL,
+		.props_size_get = &lzma_simple_props_size,
+		.props_encode = &lzma_simple_props_encode,
+	},
+#endif
 #ifdef HAVE_ENCODER_DELTA
 	{
 		.id = LZMA_FILTER_DELTA,
--- a/src/liblzma/common/string_conversion.c
+++ b/src/liblzma/common/string_conversion.c
@@ -258,7 +258,9 @@ static const char *parse_options(const char **const str, const char *str_end,
 		|| defined(HAVE_ENCODER_SPARC) \
 		|| defined(HAVE_DECODER_SPARC) \
 		|| defined(HAVE_ENCODER_RISCV) \
-		|| defined(HAVE_DECODER_RISCV)
+		|| defined(HAVE_DECODER_RISCV) \
+		|| defined(HAVE_ENCODER_LOONGARCH) \
+		|| defined(HAVE_DECODER_LOONGARCH)
 static const option_map bcj_optmap[] = {
 	{
 		.name = "start",
@@ -523,6 +525,11 @@ static const struct {
 
 #if defined(HAVE_ENCODER_RISCV) || defined(HAVE_DECODER_RISCV)
 	{ "riscv",        sizeof(lzma_options_bcj),   LZMA_FILTER_RISCV,
+	  &parse_bcj,     bcj_optmap, 1, 1, true },
+#endif
+
+#if defined(HAVE_ENCODER_LOONGARCH) || defined(HAVE_DECODER_LOONGARCH)
+	{ "loongarch",        sizeof(lzma_options_bcj),   LZMA_FILTER_LOONGARCH,
 	  &parse_bcj,     bcj_optmap, 1, 1, true },
 #endif
 
--- a/src/liblzma/liblzma_generic.map
+++ b/src/liblzma/liblzma_generic.map
@@ -135,4 +135,6 @@ global:
 	lzma_bcj_riscv_decode;
 	lzma_bcj_x86_encode;
 	lzma_bcj_x86_decode;
+	lzma_bcj_loongarch_encode;
+	lzma_bcj_loongarch_decode;
 } XZ_5.6.0;
--- a/src/liblzma/liblzma_linux.map
+++ b/src/liblzma/liblzma_linux.map
@@ -150,4 +150,6 @@ global:
 	lzma_bcj_riscv_decode;
 	lzma_bcj_x86_encode;
 	lzma_bcj_x86_decode;
+	lzma_bcj_loongarch_encode;
+	lzma_bcj_loongarch_decode;
 } XZ_5.6.0;
--- a/src/liblzma/simple/Makefile.inc
+++ b/src/liblzma/simple/Makefile.inc
@@ -49,3 +49,7 @@ endif
 if COND_FILTER_RISCV
 liblzma_la_SOURCES += simple/riscv.c
 endif
+
+if COND_FILTER_LOONGARCH
+liblzma_la_SOURCES += simple/loongarch.c
+endif
--- a/src/liblzma/simple/loongarch.c
+++ b/src/liblzma/simple/loongarch.c
@@ -1,0 +1,266 @@
+// SPDX-License-Identifier: 0BSD
+
+///////////////////////////////////////////////////////////////////////////////
+//
+/// \file       loongarch.c
+/// \brief      Filter for loongarch binaries
+///
+/// This converts loongarch relative addresses in the BL, PCALAU12I. PCADDU18I
+/// immediates to absolute values to increase redundancy of loongarch code.
+///
+/// There are many other PC-relative instrutions, including PCADDI, PCADDU12I
+/// AND B, converting them was tested but they are not very useful, most of
+/// them is not used for functions calls, even if they do, their range are
+/// usually too small so encoding them usually reduce the redundancy.
+//
+//  Authors:    Lasse Collin
+//
+///////////////////////////////////////////////////////////////////////////////
+
+#include "simple_private.h"
+
+// This macro checks multiple conditions:
+// (1) inst2 is JIRL
+// (2) JIRL's rj equals to PCADDU18I's rd
+// (3) JIRL's rd = $ra/$zero
+// These conditions could be found in laelf.adoc
+// we use them to further reduce false-positives.
+#define NOT_PCADDU18I_PAIR(pcaddu18i, jirl) \
+	(((jirl >> 26) != 0x13) \
+	|| ((pcaddu18i & 0x1F) != ((jirl >> 5) & 0x1F)) \
+	|| ((jirl & 0x1F) != 0x1 && (jirl & 0x1F) != 0x0))
+
+static size_t
+loongarch_code(void *simple_ptr, uint32_t now_pos, bool is_encoder,
+		uint8_t *buffer, size_t size)
+{
+	if (size < 12)
+		return 0;
+
+	size -= 12;
+
+	size_t i;
+
+	for (i = 0; i <= size; i += 4) {
+		uint32_t pc = (uint32_t)(now_pos + i + 4);
+		uint32_t inst = read32le(buffer + i + 4);
+
+		if ((inst >> 26) == 0x15) {
+			// BL instruction:
+			// The full 26-bit immediate is converted.
+			// The range is +/-128 MiB.
+			//
+			// Using the full range helps quite a lot with
+			// big executables. Smaller range would reduce false
+			// positives in non-code sections of the input though
+			// so this is a compromise that slightly favors big
+			// files. With the full range, only six bits of the 32
+			// need to match to trigger a conversion.
+			const uint32_t src = ((inst & 0x3FF) << 16)
+					| ((inst & 0x3FFFFFF) >> 10);
+
+			// We use previous instruction of the BL to reduce
+			// false positives while filtering non-code buffer,
+			// the previous instructions we use below are get
+			// by analyze loongarch executables and libraries.
+			uint32_t prev_inst = read32le(buffer + i);
+
+			// The instructions below are (in opcode order):
+			// 1. OR ~ LD.WU
+			// 2. BEQZ, BNEZ, B, BL
+			// Only if prev_inst are in these instructions,
+			// we'll convert BL's immediate.
+			if ((prev_inst - 0x00150000) > 0x2A800000 - 0x00150000
+				&& (prev_inst & 0xE8000000) != 0x40000000)
+				continue;
+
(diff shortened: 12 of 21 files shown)


Larhzu:
Here are some quick early comments.

You can use the `simple` pointer to safely store data between calls. See the x86 filter for an example.

If you only care about the previous four bytes, set `unfiltered_max` to `8` instead of `4` when calling `lzma_simple_coder_init`. Then look forward four bytes further, that is, instead of

```
uint32_t instr = read32le(buffer + i);
```

do this:

```
uint32_t instr = read32le(buffer + i + 4);
```

Then you have the previous instruction in `buffer + i`. Maybe you won't need to save any state then. At least this avoids the out-of-bounds access that currently can happen here when `i == 0`:

```
uint32_t prev_instr = read32le(buffer + i - 4);
```

False positives can be a problem with other filters too, but your description sounds like that there is a common pattern in debug sections that causes too much trouble. A higher level smart filter, that applies these simple filters only to code sections, was planned over 15 years ago, but it still hasn't been implemented. (Compare to 7-Zip which has done fancy things over two decades in the .7z format. It slightly helps that 7-Zip does archiving too, so it can seek in the input files and even sort them.) So, if false positives are a big problem now, then the filter perhaps has to handle them to some extent at least like you have done now.

The current list of previous instructions to match is quite long in sense that the if-statement has many conditions. This is good in the beginning, but hopefully it can be simplified to a shorter "good enough" condition. The development versions of the RISC-V filter had a few too-specific conditions too that were simplified for the final version.

I recommended avoiding official Filter ID ranges until the filter is final. That way we avoid accidents where there can be files that have official ID but cannot be decoded by stable tools. Use random 40-bit personal ID followed by 16-bit filter number, for example, 0x47BEC794C6xxxx where xxxx can be whatever you like. :-)

I'm not able to promise quick progress with this. Getting it merged this year and included in xz 5.10.0 early next year could be nice. I don't promise anything, things can happen faster or slower.

I suppose you tried tweaking LZMA2 options too. Since the instructions are 32 bits, I guess `pb=2,lp=2,lc=2` should be the best like with ARM64. If you haven't tested those yet, it's worth doing.

Thanks!

lrzlin:
Thanks for your quick reply!

> You can use the `simple` pointer to safely store data between calls. See the x86 filter for an example.

I've tried set `unfiltered_max` to `8` and look forward four bytes further, unfortunately, it not works well, caused regression and even data corrupt, so I choosed the `simple` pointer, which solved the out-of-bounds access & multithread problem perfectly, I also deleted `prev_pos` because we don't need it anymore.

> The current list of previous instructions to match is quite long in sense that the if-statement has many conditions. This is good in the beginning, but hopefully it can be simplified to a shorter "good enough" condition. The development versions of the RISC-V filter had a few too-specific conditions too that were simplified for the final version.

I tried to get rid of some conditions, but it seems that the previous instruction list shows best balance between code sections & debug sections, it even improves compress ratio for some stripped .a files, for example: `libQt6QmlLS.a`

Compress without BCJ filter
```
libQt6QmlLS.a (1/1)
  100 %        86.9 MiB / 125.1 MiB = 0.694    13 MiB/s       0:09
``` 

Compress with origin BCJ filter
```
libQt6QmlLS.a (1/1)
  100 %        87.5 MiB / 125.1 MiB = 0.699    12 MiB/s       0:10
```

Compress with improved BCJ filter
```
libQt6QmlLS.a (1/1)
  100 %        87.0 MiB / 125.1 MiB = 0.696    13 MiB/s       0:09
```
But I do agree the condition now is a bit longer, which slows down compress speed for bigger file (1.0G~), I'll discuss this with loongarch developers about the debug section, hope we'll find a simpler list in the future.

> I recommended avoiding official Filter ID ranges until the filter is final. That way we avoid accidents where there can be files that have official ID but cannot be decoded by stable tools. Use random 40-bit personal ID followed by 16-bit filter number, for example, 0x47BEC794C6xxxx where xxxx can be whatever you like. :-)

I've temporary changed the ID to `0x47BEC794C61203`.

> I'm not able to promise quick progress with this. Getting it merged this year and included in xz 5.10.0 early next year could be nice. I don't promise anything, things can happen faster or slower.

Thanks for your help, I'm not familiar with xz's release cycle & version policy, I think we could decide the version number and Filter ID as you find it's good to merge.

> I suppose you tried tweaking LZMA2 options too. Since the instructions are 32 bits, I guess `pb=2,lp=2,lc=2` should be the best like with ARM64. If you haven't tested those yet, it's worth doing.

Yes, when dealing with most executables & libraries, `pb=2,lp=2,lc=2` do performs well, as for the bigger libs with debug sections (such as the `libQt6QmlLS.a` I mentioned before, with has 1.6G in size without strip), the default options shows better compress ratio.

lrzlin:
> What you have now looks correct. :-) However, I still suspect it could be nicer without `last_ins` state, but this is a minor thing that can be done among the last steps of the filter design. I mention it now already because the topic is on my mind:
> 
> Did you also update the initial `size &= ~(size_t)3;` to
> 
> ```
> if (size < 8)
>     return 0;
> 
> size -= 8;
> ```
> 
> and the for-loop condition to `i <= size`? See riscv.c, you just want to keep `i += 4`.

Thanks for your advice! After update the `size &= ~(size_t)3;`, everything runs fine, now we could just delete `last_ins` state.

> The condition accepts about 31 of possible 256 8-bit prefixes. That's about 3 bits that have to match in addition to `(instr >> 26) == 0x15)` which requires 6 bits to match. Replacing the `prev_instr` condition with a simple one like `prev_instr <= 0x5F000000` would be weaker, providing only about 1.4 bits instead of 3. It's an example that is worth testing, but it wouldn't surprise me if it isn't good enough.
 
I've tried different conditions, the `prev_instr <= 0x5C00000` (which is the last instruction of LoongArch base instruction set) seems working well (e.g. for `libQt6QmlLS.a`, from 1.6G to 212.5MB), However, for better compress ratio, I decided to add the lower bound `0x00150000` (OR's opcode) and alter the lower bound to `0x2A800000` (LD.WU's opcode), then add B, BL, BEQZ, BNEZ seperately. In that case, we could compress `libQt6QmlLS.a` to 208.7MB, as a comparation, the condition before is 207MB, and the original one make it to 216MB, and 206MB without BCJ filter.

> If debug sections are a special trouble, I wonder what are the common bit pattern(s) that cause false matches there.
 
The develop system I used is AOSC, it didn't strip the libraries like other Linux release, so at beginning I thought that's the reason why their binaries are so large. For example, 1.6GB for libQt6QmlLS.a while other distros like LoongArchArchLinux have 99MB. However after I stripped them the AOSC's libQt6QmlLS.a is still 125.1MB but the ArchLinux one's dropped to only 5.1MB, I'll discuss this difference with AOSC developers to find out the truth.

> Release cycle is what it happens to be. Sorry if I sounded pessimistic in my previous post, I just didn't want to set too high expectations because it's hard to predict the future. When you are happy with a version of the filter, there are more steps that have to happen before the filter is included in a stable XZ Utils release.
 
> First, I need to study LoongArch64 instructions to verify the proposed implementation. I also should understand if all instructions worth filtering are being handled. For example, [there seem to be](https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN#_pcaddi_pcaddu121_pcaddu18l_pcalau12i) instructions similar to `PCALAU12I`, but I have no idea how much they are used in real-world code.

Yes, there are `PCADDI` & `PCADDU12I` which similar to `PCALAU12I` and I've tried add them to the filter, but they were relatively less used in excutables and libraries, so they really didn't help with the increasement of redundancy. Also, LoongArch compiler's mostly use `BL` and `PCALAU12I` for function calls, so at least for current compiler's, only filter two them should be good enough.

> Test files are needed to compare different filter variants. If debug symbols are a special concern, 2-3 executables or shared libraries (not static libs) are needed. If you can provide them, that would be great. Debian's LoongArch64 port seems to have [some packages](https://deb.debian.org/debian-ports/pool-loong64/main/) available, which hopefully are enough for testing non-debug files.

Yes, we could use Debian's packages for testing non-debug files. As for debug files, AOSC should be okay for testing it, they just keep all the debug info in their packages, and they also used deb files so we can easily used them for testing. Here is the package download site: https://packages.aosc.io/ 

> Once both you and I feel OK about the design, I want to discuss it with 7-Zip's developer Igor Pavlov. Test files are needed for this too.
> 
> I hope this explains a bit why this will take time.

Thanks for your kindly explain, thus I'll keep on working on it until we found it OK for merge.

Larhzu:
I still haven't tested the filter, but on the surface it looks fairly nice already. :-) However, there is more work to do still.

## Instructions to filter

> Yes, there are PCADDI & PCADDU12I which similar to PCALAU12I and I've tried add them to the filter, but they were relatively less used in excutables and libraries, so they really didn't help with the increasement of redundancy. Also, LoongArch compiler's mostly use BL and PCALAU12I for function calls, so at least for current compiler's, only filter two them should be good enough.

It would be good to understand why the instructions`PCADDI`, `PCADDU12I`, and `PCADDU18I` exist. Then one can estimate how much they can matter in the filter. Also, compiler options might affect how much they are used (for example, on RISC-V, `-fPIC` makes `AUIPC` common). If one only looks at existing binaries in one distro, one might miss something that matters in some other situation.

binutils' [elfnn-loongarch.c](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=bfd/elfnn-loongarch.c;h=faad512f3396b09fdd8106daca0cd455b5a5cff2;hb=HEAD) is one source of information.

### PCADDI

Linker relaxation of `PCALAU12I + ADDI.D` can result in `PCADDI`. See `loongarch_relax_pcala_addi` in the binutils source file.

If I understand correctly, `PCADDI` can be used to get the address of nearby (±2 MiB) functions and 4-byte-aligned global variables. *Maybe* it's somewhat common if there are many functions that refer to the same global variables, and the executable or shared library is small enough.

ARM64's `ADR` has ±1 MiB range, so it's somewhat similar to `PCADDI`. The ARM64 filter doesn't modify `ADR` instructions. So maybe `PCADDI` isn't worth filtering. False positives would be a problem too like they are with `BL`.

### PCADDU12I

An [FAQ](https://blog.xen0n.name/en/posts/tinkering/loongarch-faq/) says that `PCADDU12I` is used in PLT stubs. The binutils source file confirms this. If this was the only use of `PCADDU12I` then perhaps filtering it isn't important.

I read that 32-bit LA32R might lack `PCALAU12I` and thus such binaries use `PCADDU12I`. LA32R and also LA32S lack `PCADDU18I`, but that instruction shouldn't be useful on a 32-bit system anyway.

Full filtering of `PCADDU12I` might be like with RISC-V's `AUIPC` because the instruction can be paired with a few other instructions. Does the filter matter on LA32R? A compromise could be to filter `PCADDU12I + JIRL` which, based on RISC-V, could be the most common pairing. (Other pairs to `PCADDU12I` could be `ADDI.W/D` and loads and stores.) Or there could be separate LA32R filter later if needed, and ignore `PCADDU12I` for now.

### PCADDU18I

The [code models](https://github.com/loongson/la-abi-specs/blob/release/laelf.adoc#code-models) section in the psABI shows that `PCADDU18I` is used with the medium code model. I got an impression that LLVM recently switched to medium by default because big applications fail to link with the 128 MiB limit of the normal/small code model.

When compiling for the medium code model, object files (and thus static libs) should have many `PCADDU18I + JIRL` pairs. However, linker relaxation turns them into `BL` or `B` if the range is small enough. See `loongarch_relax_call36` in the linked binutils source file. Thus, the `PCADDU18I` might not appear in executables and shared libs unless they are big enough (over 128 MiB).

If `PCADDU18I` is filtered, it highly likely should be filtered as `PCADDU18I + JIRL` pair. The `JIRL` contains the low bits of the address. There doesn't seem to be any other instruction that would neatly pair with `PCADDU18I` to calculate an address or load/store data.

Filtering `PCADDU18I + JIRL` shouldn't cause many false matches. 6+5 opcode bits must match. Also `rj` of `JIRL` must equal `rd` of `PCADDU18I`, which provides extra five bits to the match requirement. Due to how the pseudo-instrucions [call36 and tail36](https://github.com/loongson/la-asm-manual/blob/main/adoc/CHAPTER-9.Pseudo_Instruction_Set/9.Pseudo_Instruction_Set.adoc#923-call36) are defined, `rd` of `JIRL` should normally be the same register too, and matching this provides another five bits. Then one would have 6+5+5+5 bits to match already. One could even restrict the range to something smaller than ±128 GiB. For example, ±8 GiB range would require 6+5+5+5+4=25 bits to match, but that might be stricter than required in practice.

Technically it's possible to use `PCALAU12I + JIRL` and even `PCADDU12I + JIRL` for long jumps too. The binutils file does mention `pcalau12i + jirl` in some context but maybe it's a legacy thing, I didn't investigate much.

### Other thoughts about PCAxxx

Like with RISC-V's `AUIPC`, LoongArch's `PCADDU12I` and `PCADDU18I` cannot be perfectly filtered without the paired instruction: the same absolute address can be referenced with two consecutive relative addresses in `PCADDU1xI` because the paired instruction uses a sign-extended immediate. Matching the paired instruction reduces false matches considerably too, but it only works (easily) if the instructions are adjacent.

It's good to look at output from more than one toolchain. With RISC-V, the executables from the Go compiler had some small differences in filtering results compared to GCC and Clang/LLVM, so binaries of programs written in Go might be worth checking out.

Curiously, Go's link tool uses [trampolines](https://github.com/golang/go/blob/d73aa535ded764933aa63e4c2c8fb34e479270cb/src/cmd/link/internal/loong64/asm.go#L333) (it's called [veneer](https://developer.arm.com/documentation/dui0803/c/Image-Structure-and-Generation/Linker-generated-veneers/What-is-a-veneer-) in ARM64). This suggests that one won't find any `PCADDU18I` instructions from Go binaries even if they exceeded 128 MiB.

### Summary

From my side, this is still a lot of guessing and hoping instead of actually knowing something.

* Without looking at disassembly of binaries, I don't have a confident guess about `PCADDI`, but *maybe* it doesn't need to be filtered as you had observed already.

* I suspect that `PCADDU18I + JIRL` combination is worth filtering, but one needs big (>128 MiB) executables or shared libs to find such instruction pairs.

* `PCADDU12I + some_instr` hopefully is rare and thus not worth filtering outside LA32R. If it was filtered, the paired instruction should be filtered too like with `PCADDU18I`, but now multiple instructions are possible pairs like with RISC-V's `AUIPC`.

Trying to filter also `B` in addition to `BL` is worth testing, but based on the experience from other instruction sets, the results likely are mixed or negative. `B` is used for tail calls (which should be worth filtering) and short in-function jumps (maybe not worth filtering).

* * *

## Unaligned access

From what I read, it sounds like that LoongArch processors don't necessarily support fast unaligned memory access. This is true on RISC-V and ARM64 too although I have (possibly incorrectly) an impression that most ARM64 processors have fast unaligned access. While the LoongArch instructions are four-byte aligned, the filter in liblzma has to work with unaligned buffers too.

`read32le` is safe with unaligned input, but if unaligned isn't fast, compiler may generate byte-by-byte access. Detecting `BL` and `PCALAU12I` requires checking only one byte. It might be faster to compare only one byte, and load the rest of the bytes only when the opcode matches.

Since you have access to real LoongArch hardware, it could be useful to check if there's a speed difference. One should test the filter alone, for example, have a gigabyte-sized buffer and repeatedly calling the filter/unfilter function directly on the buffer (without compression), and comparing the speed: aligned vs. unaligned buffer, and `read32le` vs. byte-by-byte reads. One shouldn't worry about this too much; there's no need to spend a lot of time on this.

* * *

## Branch conditions

In my previous post I messed up the bitmask math, sorry. I try again with the new code:

```
if ((prev_instr < 0x00150000 || prev_instr > 0x2A800000)
        && (prev_instr & 0xF8000000) != 0x40000000
        && (prev_instr & 0xF8000000) != 0x50000000)
    continue;
```

The last two conditions can be combined to `(prev_instr & 0xE8000000) != 0x40000000`. One branch is avoided if the first two are combined using subtraction, but only testing tells if it makes any kind of difference (the code only runs when the `BL` opcode matches so it might not really matter).

```
if ((uint32_t)(prev_instr - 0x00150000) > 0x2A800000 - 0x00150000
        && (prev_instr & 0xE8000000) != 0x40000000)
    continue;
```

At this point this is a *very* minor silly thing, but I mentioned it anyway because it was on my mind. :-)

* * *

## Static libs

> In that case, we could compress libQt6QmlLS.a to 208.7MB, as a comparation, the condition before is 207MB, and the original one make it to 216MB, and 206MB without BCJ filter.

> edit: After talking with AOSC developers, the size difference should be blame to LTO(Linker Time Optimization), AOSC enable it by default for better performance, however it'll generate .gnu.lto section in ELF files, which gnu strip could not deal with, we could use llvm-strip to remove these LTO sections. Though we could remove these sections, the filter still performs worse than the lzma2 itself, but better than the original filter(without conditions), which should be the common issue for all filters when dealing with .a files?

Static libraries (.a files) have nothing to filter. They consist of object files (.o) where the relative addresses are filled with zeros. If a package has a few tiny static libraries or objects (.o files) and also big executables and shared libraries, then the small .a and .o files might not matter much. But otherwise one should avoid filtering .a, .o, and .ko (Linux kernel modules).

The best solution is to make xz smarter and detect file types in the input stream. This has been planned for a very long time. I don't know when it will be implemented, but some version could be worth trying for the next release cycle (even a simple but kind of bad one, that is, working somewhat OK if input archive has the files sorted by type so all executables and shared libs are together).

It's also possible to make the filter relatively harmless on .a and .o files by implementing zero-skipping (the relative addresses are zeros in .a and .o and .ko). This was debated in length with the ARM64 filter design. In the end, it wasn't included. It's not much code, but if smarter filtering is implemented later, then the zero-skip feature has slightly negative value. So while zero-skip is technicaly simple, it kind of feels complicated at the same time. :-|

Because the stricter condition for `BL` helped significantly with .a files, the stricter condition likely is a good addition even if your testing was done with .a files. Matching `BL` alone requires only six bits to match, which isn't much. As long as the stricter condition doesn't hurt with executables and shared libraries, it is worth keeping. :-) (It even makes me wonder if the ARM64 filter could have had something like that. The false matches with `BL` are a problem there too.)

* * *

Sorry about the overlong post. The important thing is to figure out if there are more instructions or instruction pairs that should be filtered.

thesamesam:
> edit: After talking with AOSC developers, the size difference should be blame to LTO(Linker Time Optimization), AOSC enable it by default for better performance, however it'll generate .gnu.lto section in ELF files, which gnu strip could not deal with, we could use llvm-strip to remove these LTO sections. Though we could remove these sections, the filter still performs worse than the lzma2 itself, but better than the original filter(without conditions), which should be the common issue for all filters when dealing with .a files?

When you build with `-flto` and build a static library with non-ancient GCC, the resultant `foo.a` is a static archive containing only LTO IR/bytecode. If you link against `foo.a`, it will be using LTO. As @Larhzu says, it's not object code at all.

If you build with `-flto -ffat-lto-objects`, GCC will populate `foo.a` with both LTO IR sections and regular object code, and then strip can remove the LTO part.

GNU strip should support this more easily with upcoming Binutils 2.45, thanks to H.J. Lu (https://sourceware.org/PR21479). But you can just do `strip -R .gnu.lto_* -R .gnu.debuglto_* -N __gnu_lto_v1` today which works fine (you can drop the `-N ...` if you're only supporting non-ancient GCC).

Clang these days supports `-ffat-lto-objects` too. The alternative would be, for Clang and I think GCC too, is to recompile those objects using the driver to regular ELF objects.

TL;DR: don't distribute LTO'd static archives, make sure you build w/ `-ffat-lto-objects` and strip them (using the recipe above). You usually don't want LTO'd static archives because they require any consumers to have the same toolchain version.

lrzlin:
Thanks for your detailed post and @thesamesam kindly pointed the LTO related issue, I've told the AOSC developers to strip the LTO section of their static libraries, which reduce the binary size in a huge way.

### PC relative & B filtering
I've tried to add other PC relative instructions and B to the filter, here are the results:
1. `PCADDU12I` is only used for PLT entries in loongarch64 binaries, so filtering it is somewhat decline the redundency, though LA32R is lack of PCALAU12I, the only use of this subset is for educational purpose, so I think we could safely ignore it by now.
2. B does showed a mixed results, on some executables it did help, but when dealing with all binaries in `/bin`, it even reduced the compression ratio, so I think it's not worth filtering like other architectures.
3. `PCADDU18I + JIRL` shows a great improment on executables and dynamic libraries, when zero-skipping it even didn't hurt .o and .a files' compress ratio. However adding zero-skipping to other instructions filtered like `BL` or `PCALAU12I` caused data corruption, also it did help with the ratio too much, so I just add this to the `PCADDU18I` only.
4. `PCADDI`, to my surprise, maybe worth filtering, while it did increase the compressed size of .a/.o files, it shows a constent improve on binaries, also it's filtering logic is pretty simple, I'll show the data in the chart below, but compared with PCADDU18I, the false-positives is still an issue, which could not be resolved by zero-skipping due to data corruption.

About the implement of `PCADDU18I + JIRL`, due to the address could be 38-bits long, I decide to put them seperately into two instuctions, instead of using standalone address like RISC-V. Also, I did tried the big-endian address implement, but it didn't show any difference in compress ratio, so I just used the seperate way which could support full 38-bit address length.

Here are the testing results, skip stands for zero-skipping, SLib is static libraries with debuginfo, but without LTO data.
| Filtered Instructions            | Bin             | SLib            | DLib            | Striped SLib    | Overall         |
|----------------------------------|-----------------|-----------------|-----------------|-----------------|-----------------|
| None                             | 818.1MiB/0.217 | 553.9MiB/0.118 | 918.1MiB/0.229 | 276.0MiB/0.255 | 2566.1MiB/0.189 |
| Plain BL + PCALAU12I             | 787.1MiB/0.209  | 585.3MiB/0.124  | 866.9MiB/0.216  | 284.6MiB/0.263  | 2523.9MiB/0.186 |
| BL/PCALAU12I                     | 783.7MiB/0.208  | 567.9MiB/0.121  | 865.9MiB/0.216  | 280.8MiB/0.260  | 2498.3MiB/0.184 |
| BL/PCALAU12I/ADDU18I             | 757.1MiB/0.201  | 573.3MiB/0.122  | 849.0MiB/0.211  | 286.2MiB/0.265  | 2465.6MiB/0.182 |
| BL/PCALAU12I/ADDU18I (skip)      | 757.0MiB/0.201  | 567.8MiB/0.121  | 849.0MiB/0.211  | 280.8MiB/0.260  | 2454.6MiB/0.181 |
| BL/PCALAU12I/ADDU18I/ADDI (skip) | 752.9MiB/0.200  | 573.8MiB/0.122  | 842.6MiB/0.210  | 282.5MiB/0.261  | 2451.8MiB/0.181 |

Base on lastest gcc/llvm toolchain and renewed AOSC Linux, removed all LTO sections in static libraries.

### Branch Conditions & Static Libs
```
if ((uint32_t)(prev_instr - 0x00150000) > 0x2A800000 - 0x00150000
        && (prev_instr & 0xE8000000) != 0x40000000)
    continue;
```
> The last two conditions can be combined to (prev_instr & 0xE8000000) != 0x40000000. One branch is avoided if the first two are combined using subtraction, but only testing tells if it makes any kind of difference (the code only runs when the BL opcode matches so it might not really matter).

I tried new branch conditions, it did speed up the filter a bit, also through testing results, even after the LTO sections was removed, the conditions still improves redundency, so it's still worth keeping. 

### Unaligned access
Strangely enough, on LoongArch hardware (a Loongson 3A6000, 4c/8t, runs on 2.5 GHz), I couldn't find a steady relation with the filter's complexity and time consumption, when dealing with large chunks of data (like the bin in the chart which is 3.7 GiB~ large), the most complex filter (`conditional`BL` + `PCALAU12I` + `PCADDI` + `PCADDU18I/JIRL`) even runs faster than without BCJ filter (4:35 v.s. 4:53), I'm pretty confused with it, the LoongArch processor didn't support dynamic boost, is it related to process scheduling or this shows unaligned access isn't a problem we need to worry about?

### Zero-skipping and conditional BL filtering
> It's also possible to make the filter relatively harmless on .a and .o files by implementing zero-skipping (the relative addresses are zeros in .a and .o and .ko). This was debated in length with the ARM64 filter design. In the end, it wasn't included. It's not much code, but if smarter filtering is implemented later, then the zero-skip feature has slightly negative value. So while zero-skip is technicaly simple, it kind of feels complicated at the same time. :-|

After implementing `PCADDU18I + JIRL` and zero-skipping, we further improves the compress ratio of this filter, due to zero-skipping for this instruction pair did not show other side effects, so I'm curious about the negative value about zero-skipping, does it makes the smart filtering harder to write? I think if we're going to do so, the conditional `BL` and zero-skipping of `PCADDU18I + JIRL` could be useless in some way and should be removed, but they did make a better compress ratio by far.

Sorry for the late reply, it spend my many time to figure out the best way to implement `PCADDU18I + JIRL` pair filtering, hopefully this makes progress to our final results.

Larhzu:
Sorry for the delay. I was on an unannounced long break. Before the break I had drafted half a response, but I need to re-read things to remember the details and finish the reply. Hopefully I get it done next week.

lrzlin:
@Larhzu Gentle Ping. Any comments for my last reply?

Larhzu:
Sorry for the silence. I quickly tried the code three months ago and also a few variations, and the results looked good with your version. I wasn't able to focus on the code more though, so there's no progress.

I had read your latest reply back then and I read it again now. It sounds good that the same check for previous instruction works also with `PCADDI`. I don't have new thoughts about zero skip.

With desktop CPUs, it makes sense that there's no practical difference between aligned and unaligned speed. If there's a problem on embedded processors, the code can be changed later in a compatible manner.

`now_pos` is `uint32_t` so if distance is huge with `PCADDU18I`, the conversion won't be done quite correctly: it won't corrupt data, but compression might be a tiny bit worse with the distant addresses. The downside might not matter in practice; keeping the code simpler may have more value. I'd like to avoid changing `now_pos` to 64 bits because one needs to change the other filters to handle it correctly (in liblzma and in other implementations).

I currently feel that a new stable release with new features (5.10.0) is at least six months away. The LoongArch filter will be included in it. I need to think about the code, test it a lot more, and so on. I will reply here when I have news.

PR #177, opened 2025-05-07
xzgrep: Support `-r` / `--recursive`

KSR-Yasuda:
Support `-r` / `--recursive` option.

If the option is set, replace directory paths into descendent file paths in advance.

Uses the external command `find`.
Is it acceptable?

----

I know I can run as
`xzgrep "pattern" $(find . -type f)` or `find . -type f | xargs -I{} xzgrep "pattern" {}`
without `-r` option.

However, to use `xzgrep` as Vim `&grepprg`, it is preferable to handle directories by itself.

### DIFF
--- a/src/scripts/xzgrep.1
+++ b/src/scripts/xzgrep.1
@@ -73,9 +73,6 @@ of
 are supported.
 However, the following options are not supported:
 .IP "" 4
-.BR \-r ,
-.B \-\-recursive
-.IP "" 4
 .BR \-R ,
 .B \-\-dereference\-recursive
 .IP "" 4
--- a/src/scripts/xzgrep.in
+++ b/src/scripts/xzgrep.in
@@ -57,6 +57,7 @@ files_with_matches=0
 files_without_matches=0
 no_filename=0
 with_filename=0
+recursive=0
 
 # See if -H and --label options are supported (GNU and *BSDs).
 if test f:x = "$(eval "echo x | $grep -H --label=f x 2> /dev/null")"; then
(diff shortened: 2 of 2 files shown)


KSR-Yasuda:
Also, it had better allow `-r` option right before directory path
(Now options must be given before search pattern).

Tried, but I didn't see how it comes clearly.

Larhzu:
Adding `--recursive` could be nice, but shell scripts can be tricky to write correctly. There are also portability concerns.

1. Arrays are a bash feature that isn't available in POSIX shell.

2. The `while read line` loop doesn't handle special characters in filenames. For example, filenames with newline (`\n`) characters or backslashes will be mishandled.

   Using `read -r` fixes the backslash handling, and `-d ''` fixes newlines by using a null character to terminate the string (one then needs `-print0` with `find`). The `-r` option should be portable enough. The `-d` option was only added in POSIX.1-2024 and doesn't seem so widely supported yet, and thus bash might be required.

   Leading and trailing white space is still removed. This can be fixed by setting `IFS=`, that is: `while IFS= read -r -d '' line`

3. `find -mindepth` is supported on GNU and *BSDs so it should be OK even though it's not in POSIX. However, as I understand it, there's no need to use that option here.

4. Redirection from a command (like `< <(find ...)`) is a bash feature that isn't in POSIX.

5. The `-a` and `-o` options of `test` are obsolete and shouldn't be used in portable scripts. They were removed in POSIX.1-2024. Use: `if test $recursive -ne 0 && test -d "$i"`

I don't have an opinion yet if requiring bash is OK for `--recursive`. It's not a showstopper if there is no other way, but it will require build system changes to ensure that xzgrep is run with bash.

I'm not sure if it is a good idea to read the full filename list in memory before starting to grep the files. A huge list consumes memory and also delays the start of the actual grepping. Perhaps it would be better to grep each file as their names are produced by `find`.

If someone else has thoughts about this, please comment.

A very different approach would be to patch GNU grep to use libarchive. Or improve some other grep implementation that already supports compressed files. For example, `pcregrep` supports `.gz` and `.bz2` files but not other formats.

PR #181, opened 2025-05-21
Release resources even for non-debug builds.

piru:
This change makes the applications actually free the associated resources even for non-debug builds.

Not all platforms implement shared libraries in the way that the library allocator links to the parent application's heap. Not releasing resources can thus lead to actual memory leaks, which can be hundreds of megabytes in case of xz tools.

### DIFF
--- a/src/xz/args.c
+++ b/src/xz/args.c
@@ -905,11 +905,9 @@ args_parse(args_info *args, int argc, char **argv)
 }
 
 
-#ifndef NDEBUG
 extern void
 args_free(void)
 {
 	free(opt_block_list);
 	return;
 }
-#endif
--- a/src/xz/coder.c
+++ b/src/xz/coder.c
@@ -1456,7 +1456,6 @@ coder_run(const char *filename)
 }
 
 
-#ifndef NDEBUG
 extern void
 coder_free(void)
 {
@@ -1473,4 +1472,3 @@ coder_free(void)
 	lzma_end(&strm);
 	return;
 }
-#endif
--- a/src/xz/coder.h
+++ b/src/xz/coder.h
@@ -95,10 +95,8 @@ extern void coder_set_compression_settings(void);
 /// Compress or decompress the given file
 extern void coder_run(const char *filename);
 
-#ifndef NDEBUG
 /// Free the memory allocated for the coder and kill the worker threads.
 extern void coder_free(void);
-#endif
 
 /// Create filter chain from string
 extern void coder_add_filters_from_str(const char *filter_str);
--- a/src/xz/main.c
+++ b/src/xz/main.c
@@ -339,10 +339,8 @@ main(int argc, char **argv)
 	}
 #endif
 
-#ifndef NDEBUG
 	coder_free();
 	args_free();
-#endif
 
 	// If we have got a signal, raise it to kill the program instead
 	// of calling tuklib_exit().
--- a/src/xzdec/xzdec.c
+++ b/src/xzdec/xzdec.c
@@ -479,11 +479,7 @@ main(int argc, char **argv)
 		} while (++optind < argc);
 	}
 
-#ifndef NDEBUG
-	// Free the memory only when debugging. Freeing wastes some time,
-	// but allows detecting possible memory leaks with Valgrind.
 	lzma_end(&strm);
-#endif
 
 	tuklib_exit(EXIT_SUCCESS, EXIT_FAILURE, display_errors);
 }


thesamesam:
Can you make the commit message match the PR description please? We'd like the information to be kept in `git log` in case anything happens to GH. Thanks.

Larhzu:
Thanks for the report! This is the first time I hear about an operating system where memory allocated by an application isn't freed on exit. I'm not surprised that such platforms exist, I just haven't developed for such OSes myself. I haven't personally tested every platform on which xz runs; I rely on reports by other people.

Those frees for debug builds were added 14 years ago in 2011. Earlier it didn't free at all, which probably was a bit annoying when using Valgrind. (It's been a long time so I don't remember the details.) I do know the freeing doesn't take that many processor cycles to do, but the young me liked to avoid such useless things a bit more than was perhaps sensible. It reduced code size a few bytes too, but that too doesn't matter in a code base that is bloated anyway.

Your commit only does the most obvious thing by removing the ifndefs. It doesn't attempt to always free all allocations at exit. The remaining leaks aren't too annoying with Valgrind because they only occur with certain command line options or in some error conditions. But if you have a platform where all resources must be freed every time, then those less common cases must be handled too.

It would have been nice if you had mentioned which OS or environment is affected by this issue. Perhaps you could provide some more details to understand the topic better. [Your post](https://infosec.exchange/@harrysintonen/114546117774057851) was linked to me on IRC, and people commented that it's about AROS. Can you confirm this? (By the way, some might find colorful language the norm on social media, but a more constructive writing style would be appreciated even if one's GitHub username is literally the "devil" in Finnish.)

I want to understand these about the affected platform:

1. Is it OK that a shared library allocates memory (`xz -> liblzma -> malloc`) and the application frees it (`xz -> free`), or vice versa? I know it can be a problem on Windows if the DLL uses a different CRT than the application (same issue would happen with file descriptors in that situation).

2. Is it enough to free the memory allocated in a shared library, and OK to leak the memory allocated by the application itself? This could make the fix simpler, thus the question.

3. Are file descriptor leaks a problem? The command line tools ensure that the file descriptors 0, 1, and 2 are open by using a [helper function](https://github.com/tukaani-project/xz/blob/master/src/common/tuklib_open_stdxxx.c). If they are always open on your platform, then the helper function does nothing, and thus it doesn't create a file descriptor leak either.

4. Is the memory leak still a problem if the command line tools are linked against static liblzma? Because the leaks are in the command line tools (not liblzma), I wonder if it could be a usable workaround to link xz against static liblzma while the other apps would still link against shared liblzma.

On IRC, people commented that it's *extremely* common that apps don't free memory on exit. Yes, it can be annoying if one wants to detect actual memory leaks with Valgrind or sanitizers. Otherwise there's no harm because most apps are written for OSes where memory is freed when the process terminates. I see that, in the replies to your post, some think it's dumb and perhaps it is so, because it makes the code not portable to atypical platforms where the assumption doesn't hold.

For example, xz doesn't free anything in some error conditions: it's simpler to just print a message and exit without adding more code to free all resources that (most) OSes free at exit anyway. (With a quick try with Valgrind, bzip2 and lz4 don't seem free everything, while gzip and zstd don't at least easily leak anything.) Note that in error conditions, xz is (of course) careful to not leave incomplete files behind etc. because those the OS won't cleanup on process exit.

Anyway, I hope to understand the issue better specifically on the platform(s) relevant to you. I can then look into what is needed to fix the issue on that platform, or even how much extra code would be needed to free all allocations in all situations.

piru:
Thank you for the interest. I'll try to clarify the things as much as possible. To start: This is definitely a minor issue that affects a very niche platform. So there is really no need to take any action here. If necessary, I'll keep a local patch in place if the code cannot be changed upstream, and this is totally acceptable as well.

This is not AROS, but it's close; this is MorphOS (another AmigaOS-related OS). The platform we have doesn't have true "shared objects". That is, while we can load libraries in memory with symbolic objects, they do not get symbols referenced back to the application that loaded them. In short: This means that, for example, malloc from within the library will be separate from the application heap. So while application-performed mallocs will get released, they are disconnected from the malloc/free performed by the library. (Technically, the library system that is in place is not "shared object" in a sense that is understood by most, but it is analogous, and in this context behaves in the way described, that is: it doesn't automatically link back to the calling app).

So indeed, as you've correctly identified, 1) will be a problem. 

1) Generally this cannot be done. In case of liblzma we've taken care to handle situations where memory is released by the app by passing the app-side allocator functions (and thus reference to app-side heap) as callbacks when needed (xz/coder.c `lzma_properties_decode` call is one such use we had to carefully adjust). So we don't have a problem in this sense, but we need to carefully inspect existing code for such situations and add the necessary patches to handle them. It is a bit cumbersome, but if such use isn't super common, it is manageable.

2) Yes. This is entirely fine. Everything would have been fine if `xz` and `xzdec` would have just called the `lzma_end`. This escaped me since it was in such a `#if-block` that made it less obvious what was going on (esp since I looked at xz and not xzdec. xzdec actually documents why it was done.)

3) This is non-issue here. More generally: stdio file handles opened inside a library must be released inside the library, nor can you pass stdio file handles from the app to the library or vice versa. They're disconnected similar to the heap. In some situations (for some other libraries) we've moved stdio dependent parts to a static library stub, keeping it on the app side of things. It's not great, but sometimes the easiest solution. Some libraries have a design that allows abstracting the stdio functionality, such as OpenSSL and their BIO. With OpenSSL we could get away with tiny stubs that wrap the app side stdio to BIO which is then used with the library.

4) Linking against static liblzma has no such issues as they then share the heap, stdio and so forth. However, we created the shared library specifically to avoid large binaries and memory footprint from repeating the same code again and again.

So in the end the issue arises from the fact that the memory allocated by the library is separate from the app memory. Thus if the application doesn't properly call `lzma_end`, it will leak the memory. There's also one way to handle memory in a way that will work, but is resigning to bad design: Have per-app memory pool you can allocate / free memory from. This pool will then be released once the app terminates. However, this is ugly design in my opinion and kind of validates bad design. I'm a firm believer that something went wrong if you have to depend on such hacks to avoid memory leaks.

> On IRC, people commented that it's extremely common that apps don't free memory on exit.

They're right, and this is totally fine. However, this only holds true as long as the memory is allocated by the standard library allocator. On certain platforms (such as ours) the memory (and other resources) managed by a shared library are disconnected from the application and its standard library. This is why we need to carefully inspect the library we're about to port for their API design: Does it handle memory sensibly or assume library/app memory can be mixed? Does it require passing app side resources such as file handles or sockets? We might avoid creation of a shared library if the API is messy, and keep linking statically if the library is unsuitable.

piru:
Addressing this non-technical part separately.

> By the way, some might find colorful language the norm on social media, but a more constructive writing style would be appreciated

Some background, that explains (but of course doesn't justify) the tone and style of that post: I spent some time tracking a memory leak only to eventually hone in to this particular issue. When I found out the leak was "intentional" and justified by "performance" I was particularly ticked off.

Looking back this was unfounded. In the end this is our own fault, it is our platform that has these specific quirks. It is quite silly to expect all code to cater for the particular weirdness in our platform. I went to create this PR in some (hopefully) more constructive feedback.

I have nothing but admiration for your great work on this project.

collinfunk:
Hi all, interesting thread. I know Lasse already mentioned that not freeing memory before exit is very common. For a more concrete example, this is the recommended behavior by the GNU Coding Standards [1]:
```
Memory analysis tools such as valgrind can be useful, but don’t complicate a program merely to avoid their false alarms. For example, if memory is used until just before a process exits, don’t free it simply to silence such a tool. 
```
So most GNU programs follow that recommendation:
```
$ valgrind ls 2>&1 | grep 'still reachable'
==176436==    still reachable: 20,269 bytes in 6 blocks
$ valgrind wget gnu.org 2>&1 | grep 'still reachable'
==186743==    still reachable: 1,878 bytes in 34 blocks
```
Not to say that this is the only way to do things, but it is common practice.

[1] https://www.gnu.org/prep/standards/standards.html#Memory-Usage

Larhzu:
We discussed this on IRC on Thursday.

Now that I understand the topic better, I also understand the feelings behind the issue. I likely would have been somewhat furious for a few moments in that situation when finding the cause of the issue after a long debugging session. So, I overreacted a bit in my previous message, I'm sorry.

* * *

The point (1) is comparable to what happens on Windows if a shared library and application are linked against different CRT. In that case, memory allocated in the shared library cannot be freed by the application or vice versa. The affected liblzma functions are (hopefully) listed in [liblzma-crt-mixing.txt](https://github.com/tukaani-project/xz/blob/master/windows/liblzma-crt-mixing.txt). Clearly it's not good API design if one cares about portability, but I had no idea about this problem at the time. Perhaps liblzma could provide allocation and freeing functions in the API that call `malloc` and `free`, but I feel it's a bit late to add them now.

There are a few places where the library-allocates-but-app-frees (or vice versa) is done in `xz` and one thing also in `lzmainfo`. It's fine on Windows because the assumption is that the command line tools are linked against the same CRT as liblzma. But on MorphOS it is a problem. I think the easiest fix/workaround/hack is the rarely-used `lzma_allocator` feature. With that, liblzma can be told to use allocation functions from the command line tool's libc.

A partial alternative is ensuring that `lzma_end` is always called. This either needs care to ensure that all relevant code paths indeed call it, or using `atexit` to call it. For example, with the patch from this PR,

```
xzdec goodfile.xz notfound.xz
```

decompresses the first file, but the missing second file ends up calling `exit(EXIT_FAILURE)` without calling `lzma_end`. Thus, the commit in this PR only fixes the memory leak in xzdec on MorphOS when no errors occur. The `atexit` or `lzma_allocator` hacks would work better, and it would be more obvious that one hasn't missed some code path that doesn't call `lzma_end`.

I understood that xz threading support is disabled because MorphOS is not a multiprocessor OS. This means that memory allocations are the only resource to free from liblzma. With threads, there would also be threads to terminate and possibly destroying mutexes and condition variables, and then the `lzma_allocator` hack wouldn't work to avoid leaks.

I understood that there are no signals on MorphOS, and that it's not possible to interrupt or terminate the process in a graceful manner. So there's no need to worry about memory handling in this situation. I guess you have had to patch something in xz to disable the signal handling code. Perhaps those changes could be included in upstream code.

> There's also one way to handle memory in a way that will work, but is resigning to bad design: Have per-app memory pool you can allocate / free memory from. This pool will then be released once the app terminates. However, this is ugly design in my opinion and kind of validates bad design.

I suppose using `lzma_allocator` to redirect liblzma to use application's `malloc` and `free` is one kind of per-app memory pool. Yes, it's not pretty.

> I'm a firm believer that something went wrong if you have to depend on such hacks to avoid memory leaks.

In 2025, it could be argued that a non-embedded OS should not leak memory if an application is buggy or terminates abnormally. I guess that quite a few apps need changes to port them to an OS that can leak memory this way.

It makes perfect sense why things were done the way they were done on resource-constrained systems long ago, but when RAM sizes increased, things could be done better. For example, *if* I understand correctly, this memory leak would happen on Windows 3.1 because both code and data of a DLL is shared across processes, but it's not a problem on Windows 95 anymore.

About the overhead of `lzma_end`, it obviously isn't significant to just free the memory allocations; there aren't that many of them anyway. With threaded encoder or decoder, `lzma_end` also ends up telling the worker threads to exit, and then waits for them exit. Idle threads react quickly, and busy threads poll the state change fairly often, so it still shouldn't matter much. However, I'm not sure if there's much point to change that code now; the changes from this PR alone wouldn't fully fix the problems on MorphOS anyway.

* * *

The quote from the GNU Coding Standards sounds reasonable, and xz behaves like that in several places. However, with non-trivial data structures, memory leak detection is valuable at least in debug builds.

* * *

I created a `morphos` branch. It adds the `lzma_allocator` hack. If I didn't make mistakes, it should fix all leaks and all library-allocates-but-app-frees cases.

Larhzu:
I rebased the `morphos` branch to the latest `master`. @piru, have you been able to test it or can you test it in the near future? I would like to know if it is worth merging. It should fix the problem in all command line tools in XZ Utils and in all error conditions too.

PR #186, opened 2025-07-03
tuklib_integer: Autodetect support for unaligned access on loongarch.

Zeno-sole:
-mstrict-align is also supported on loongarch architecture

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111555
https://github.com/llvm/llvm-project/pull/85350
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/arch/loongarch/Kconfig?h=v6.12.35#n536

### DIFF
--- a/m4/tuklib_integer.m4
+++ b/m4/tuklib_integer.m4
@@ -78,7 +78,7 @@ if test "x$enable_unaligned_access" = xauto ; then
 		i?86|x86_64|powerpc|powerpc64|powerpc64le)
 			enable_unaligned_access=yes
 			;;
-		arm*|aarch64*|riscv*)
+		arm*|aarch64*|riscv*|loongarch*)
 			# On 32-bit and 64-bit ARM, GCC and Clang
 			# #define __ARM_FEATURE_UNALIGNED if
 			# unaligned access is supported.
@@ -101,7 +101,8 @@ if test "x$enable_unaligned_access" = xauto ; then
 			AC_COMPILE_IFELSE([AC_LANG_SOURCE([
 #if !defined(__ARM_FEATURE_UNALIGNED) \
 		&& !defined(__riscv_misaligned_fast) \
-		&& !defined(_MSC_VER)
+		&& !defined(_MSC_VER) \
+		&& !defined(__loongarch__) 
 compile error
 #endif
 int main(void) { return 0; }


Zeno-sole:
Could you please help review the code? @Larhzu

Larhzu:
> On the embedded use cases the kernel emulates an unaligned access from user-space if the HW does not support to do so (with an exception handler). Thus they'd still work but slower.

*From point of view of xz*, the emulation is likely worse than useless. For example, earlier this year, one person tested force-enabling unaligned access on a RISC-V processor that doesn't support HW unaligned access. It made compression 22 times slower. It wasn't as bad as I had expected, but it still makes such an xz executable useless in practice. `SIGBUS` would make it obvious that one needs to fix something. I don't expect the situation to be significantly better on LoongArch or any other arch.

There is also a half-way forcing of unaligned access: Using `--enable-unaligned-access` but compiler still being in `-mstrict-align` mode. Then the executable won't rely on kernel emulation because the unaligned access is done with `memcpy` by default. In this case the same tester saw "only" 5 to 20 percent slower compression. Note that such executables have this slowdown also on systems that support fast unaligned access in hardware. This amount of slowdown isn't obvious enough, so it can go undiagnosed for a long time.

(One can fully force unaligned access while in `-mstrict-align` mode by using the `configure` option `--enable-unsafe-type-punning` and adding `-fno-strict-aliasing` to `CFLAGS`. But this is not recommended.)

In short, `--enable-unaligned-access` should be enabled only if `-mno-strict-align` is in effect too. I think it wouldn't be acceptable for `configure` to add `-mno-strict-align` automatically; it has to be the other way around, `configure` detecting if `-mno-strict-align` is already set.

> But the kernel itself needs to be built with -mstrict-align if we want it to run on such a HW because it cannot emulate unaligned access for itself.

If the kernel is intentionally built to support hardware without unaligned access, I suppose xz (and the whole userspace) should be built that way too. Otherwise kernel is slower than needed on desktop hardware and userspace is slower (or even unusable in case of xz) on embedded hardware.

It's very much possible that I'm missing something important here.

By the way, I realized that my idea of a build-time check cannot work because one can do a native build on a more capable system than what one is targeting. So autodetection would need to observe compiler behavior, for example, compile a test program and see if it uses unaligned access:

```c
#include <string.h>

void test_unaligned(void *a, void *b)
{
    memcpy(a, b, 4);
}
```

```sh
gcc -O2 -S test_unaligned.c
sed -n '/^test_unaligned:$/,/[[:blank:]]jr[[:blank:]]*\$r1/p' test_unaligned.s \
        | grep -qFw ldptr.w \
    && echo 'Using -mno-strict-align'
```

I don't know much about LoongArch assembly but I guess `ldptr.w` isn't the only possibility here. I suppose it could be `ld.w` too. I used [this Godbolt instance](https://godbolt.org/z/Yq6o58rno) to test.

This would be quite a hack. I'm not sure if something like this is acceptable in xz. Maybe it's better to wait until GCC and Clang provide a predefined macro.

Larhzu:
For fun, I pushed a draft version that hopefully detects `-mstrict-align` using `objdump` and `grep`. It's in the `loongarch_unaligned2` branch. It's not pretty but if there's no better way and it's not OK to always enable unaligned by default, *maybe* this method is worth considering, if it works.

Zeno-sole:
> Thank you for testing and for the easy-to-read table! The results are very different than what I expected. I'm not sure right now what to think. Did the machine have other load at the same time? I'm asking because the Real and User times don't correlate very well with each other. It can be seen from the CPU Usage too. While Real Time shows (4) as the winner, User Time suggests (3) to be the winner. Normally I would expect both Times to point to the same winner. If unaligned access isn't very fast, it makes sense that (3) is the best. Even if supported in hardware, unaligned access could be somewhat slow. If (4) truly was the best, then something really weird would be going on. I would have no idea what the explanation could be. Maybe it would be better to use single-threaded mode with `time xz -c -9 -T1 testfile > /dev/null` (adding `-T1`). The test file should be smaller though to keep the testing time reasonable. If compression takes about 60 seconds, it should show the differences very well already (even 10-15 seconds might do). The test file needs to be something that compresses well but not extremely well (so /dev/zero doesn't work). Truncating Linux source tarball to smaller size or using some other smaller file will work fine. Choosing the best or second best time should work well enough. If there are outliers, calculating average might be skew the results. Don't stress about the details too much. :-) If there are differences between the build options, they should be clear enough without much analysis, I hope.

I ran the test task while the machine was idle, so this result is indeed a bit strange. I will retest it using your method.

Zeno-sole:
> > Thanks! Single-threaded mode made the results consistent between Real Time and User Time, so this was an improvement. All four tests have identical results (the tiny differences are noise). This doesn't make sense to me; at least (4) should be worse than the rest. I wonder if something has gone wrong. If the package was built with shared liblzma, running src/xz/xz should use the correct shared library. If one uses "make DESTDIR=/foo/bar install" and then runs /foo/bar/bin/xz, it will use liblzma from the system, not the one from /foo/bar/lib. To prevent all problems related to shared liblzma, pass --disable-shared to configure. Then liblzma will be statically linked into src/xz/xz, and you can also verify that the src/xz/xz files produced by the different build configurations aren't identical. If the test file is incompressible, then the speed difference might be very small. But I'm pretty sure you didn't pick such a test file, I'm just mentioning this for completeness. It's also possible that what I expect to see just isn't what the reality is. But please test --disable-shared if you didn't already use that. I don't have a better guess now, sorry.
> 
> Yes, I didn't use the `--disable-shared` parameter. I'll add it later when building the test.

Unfortunately, adding `--disable-shared` didn't make any difference; I don't know where the problem lies.

### 🔹 Test Results Summary (The best result is obtained from six tests.)

| Config | `unaligned access` | `CFLAGS` option         | Avg Real Time (s) | Avg User Time (s) | Avg Sys Time (s) | Avg CPU Usage (%) |
| :----- | :----------------: | :---------------------- | :---------------: | :---------------: | :--------------: | :---------------: |
| **1** |       enable       | `-O2 -mno-strict-align` |      63.347      |      62.700      |      0.403      |       99.61       |
| **2** |       disable      | `-O2 -mno-strict-align` |      63.339      |      62.652      |      0.454      |       99.63       |
| **3** |       disable      | `-O2 -mstrict-align`    |      63.859      |      63.089      |      0..472     |       99.59       |
| **4** |       enable       | `-O2 -mstrict-align`    |    63.888    |      63.117    |      0.514      |     99.60    |

Larhzu:
Excellent! Now the results match my expectations.

On processors that support unaligned access, compression is currently 20 % slower than it needs to be. On the other hand, if unaligned access is enabled unconditionally on LoongArch, toolchains that use -mstrict-align by default will then build a compressor that is 60 % slower than it needs to be.

I understood that it makes sense to prioritize server/desktop processors, but I still dislike the possibility of (4) happening to some people. 60 % slowdown is a lot, but it might not be enough to make people think that it's a build configuration issue. It would be nice if autodetection could handle this.

I updated the loongarch_unaligned2 branch a little. Can you test if it works? It should detect if -mstrict-align or -mno-strict-align is in effect. Since GCC's default is -mno-strict-align, running configure without arguments should produce this line:

checking if unaligned memory access should be used... yes

If you run ../configure CFLAGS="-O2 -mstrict-align" then it should say "no".

Zeno-sole:
> Thanks! There are at least two issues. 1. I had forgot to include the .c file for the CMake check. I have pushed it now. 2. Because CMake didn't fail due to the missing file, it means that it never tried to use it. It has to be due to the line elseif(PROCESSOR MATCHES "^loongarch") not detecting the architecture. The PROCESSOR variable is a lower-cased version of CMAKE_SYSTEM_PROCESSOR, which I had expected to be "loongarch64" because, to my understanding, that's what Linux "uname -m" prints. I pushed a debug commit that makes CMake print something like this:
> […](#)
> -- DEBUG CMAKE_SYSTEM_PROCESSOR="x86_64" PROCESSOR="x86_64" I would like to know what it prints on your LoongArch system.

CMake Warning:
  Ignoring extra path from command line:                                            
                                                                                    
   "/tmp/xz/xz-utils-5.8.1/build-cmake/CMAKE_SYSTEM_PROCESSOR=x86_64"               
                                                                                    
                                                                                    
CMake Warning:
  Ignoring extra path from command line:                                            
                                                                                    
   "PROCESSOR=x86_64"

PR #187, opened 2025-07-13
Start Mooving to rust

mreouven:


thesamesam:
I don't think we're interested in this, sorry. Such a thing should really be discussed before a large PR.

PR #194, opened 2025-08-26
Update GPLv2 and LGPLv2 copies from gnu.org.

jas4711:
Hi!  These were copied from gnulib: https://cgit.git.savannah.gnu.org/cgit/gnulib.git/tree/doc

### DIFF
--- a/COPYING.GPLv2
+++ b/COPYING.GPLv2
@@ -2,7 +2,7 @@
                        Version 2, June 1991
 
  Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
- 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ <https://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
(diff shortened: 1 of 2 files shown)


Larhzu:
Sorry for the delay. I was on an unannounced long break.

I managed to download (wget) gpl-3.0.txt from www.gnu.org a day or two ago but trying to download gpl-2.0.txt failed, and www.gnu.org remains down still at least from my IP address. I would have liked to verify the GPLv2 and LGPLv2.1 copies from the primary source. People on IRC reported that gnu.org works for them, so perhaps the attempt to fetch the old license was seen as bot activity. Indeed, a few people tried the wget method too and also got blocked from www.gnu.org for using GNU wget. I know the abusive bots are a major problem nowadays, so it's a tricky balance for hosting providers. :-(

The copies in Gnulib indeed match the ones in this PR, and the files in /gnu/Licenses/ on ftp.gnu.org mirrors do too. :-)

Thanks!

Issue #195, opened 2025-08-31
Do no respect memory limits which leads to `Cannot allocate memory` when unpack huge archive.


catap:
I can reproduce it by the command:
```
xz -T64 -d < /build/current/usr/ports/distfiles/chromium-139.0.7258.127.tar.xz | tar -xf -
```
which fails as:
```
xz: (stdin): Cannot allocate memory
tar: End of archive volume 1 reached
```

If I reduce number of used threads it works. If I limit memory consumption via `XZ_DEFAULTS=--memlimit=150MiB` it works!

I think that xz should limit itself by available ulimits.

Am I wrong?

Larhzu:
I don't know much about cgroups, but if I understand correctly, with cgroups v2 one would need to first read /proc/self/cgroup and then read the limit from a file or files under /sys/fs/cgroup. I suppose it isn't necessarily available in sandboxed environments, but I don't know if that's the only issue with sandboxes. A few extra syscalls at startup doesn't sound too bad because typically the (de)compression phase consumes most of the process time. In any case, reading `memory.max` sounds more complicated than checking `RLIMIT_DATA`.

A side note: There is also a-few-years-old suggestion of using `MemAvailable` from /proc/meminfo instead of total amount of RAM to determine how much memory is hopefully OK for xz to use. Perhaps it would be better than 1/4 of RAM, but on the other hand it can make xz's thread count vary in a somewhat unpredictable way (depends on the system state at xz startup). The code is in the `memavail` branch. Using `MemAvailable` wouldn't help with `RLIMIT_DATA` or cgroup limits though.

The root of the problem is that currently xz tries guess how much memory can be allocated, and if the guess is too high and allocation fails, xz simply fails. It has been a problem in the default configuration since 5.5.1alpha which enabled the threaded mode by default with a guess that 1/4 of total RAM is safe to use. Checking `MemAvailable` or `RLIMIT_DATA` or other limits could make it more likely that xz won't guess too high, but it's still a hack. The 1/4 of RAM has been simple code, but when the guess is wrong, people need to manually specify a lower memory usage limit for xz (`XZ_DEFAULTS` is a good way to do that).

[...]

PR #196, opened 2025-09-06
Use RLIMIT_DATA when it's available as default memory limit

catap:
Closes https://github.com/tukaani-project/xz/issues/195

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -335,6 +335,14 @@ if(HAVE_CLOCK_GETTIME OR HAVE_CLOCK_GETTIME_LIBRT)
     tuklib_add_definition_if(ALL HAVE_CLOCK_MONOTONIC)
 endif()
 
+check_symbol_exists(getrlimit sys/resource.h HAVE_GETRLIMIT)
+tuklib_add_definition_if(ALL HAVE_GETRLIMIT)
+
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -335,6 +335,9 @@ if(HAVE_CLOCK_GETTIME OR HAVE_CLOCK_GETTIME_LIBRT)
     tuklib_add_definition_if(ALL HAVE_CLOCK_MONOTONIC)
 endif()
 
+check_symbol_exists(getrlimit sys/resource.h HAVE_GETRLIMIT)
+check_symbol_exists(RLIMIT_DATA sys/resource.h HAVE_DECL_RLIMIT_DATA)
+check_symbol_exists(RLIM_INFINITY sys/resource.h HAVE_DECL_RLIM_INFINITY)
```
Larhzu, line 341: These set CMake variables like `HAVE_GETRLIMIT`. To create a C preprocessor macro, something like

```
if(HAVE_GETRLIMIT)
    add_compile_definitions(HAVE_GETRLIMIT)
endif()
```

is needed in addition. Or in this project, one can use a shorthand:

```
tuklib_add_definition_if(ALL HAVE_GETRLIMIT)
```

PR #204, opened 2025-12-30
Enable x86 assembler support on GNU/Hurd by default

pinotree:
It has been enabled in Debian GNU/Hurd for 15+ years using `--enable-assembler=x86`. Since it works without issues, let's enable it by default also in the upstream sources, regardless of the distro.


collinfunk:
Patch looks good to me.

You already mentioned it has been enabled on Debian GNU/Hurd, but just wanted to share the source of that here [1].

[1] https://salsa.debian.org/debian/xz-utils/-/blob/debian/unstable/debian/rules?ref_type=heads#L91

PR #209, opened 2026-02-19
CMake: Don't fail if sandbox is "auto"

bansan85:
If sandbox is set to "auto", it means that CMake will make its best effort to enable sandbox. But CMake should not fail to configure if it can't enable sandbox.

Related to https://github.com/microsoft/vcpkg/issues/49986

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1919,9 +1919,6 @@ if(NOT SANDBOX_FOUND AND XZ_SANDBOX MATCHES "^auto$|^landlock$")
         HAVE_LINUX_LANDLOCK)
 
     if(HAVE_LINUX_LANDLOCK)
-        set(SANDBOX_COMPILE_DEFINITION "HAVE_LINUX_LANDLOCK")
-        set(SANDBOX_FOUND ON)
-
         # Of our three sandbox methods, only Landlock is incompatible
         # with -fsanitize. FreeBSD 13.2 with Capsicum was tested with
         # -fsanitize=address,undefined and had no issues. OpenBSD (as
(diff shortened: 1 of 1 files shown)


Larhzu:
I understand that it might be annoying behavior, but the solution proposed in this PR isn't the way to go.

If one adds sanitizer options to CFLAGS without changing anything else, I think the assumption is that nothing else changes. If sanitizer flags silently disabled sandbox, I would find it sneaky behavior. If the sanitizer flag detection had a bug, the sandbox might become disabled even when it shouldn't be, and no one would notice.

A fair bit of thinking and discussion occurred before 88588b1246d8 was committed in late 2023. I think the commit is still good, and "auto" should behave like it does now.

I'm thinking if adding a new option "auto-unless-sanitizer" could be worth it. Then the packager would explicitly request the slightly riskier behavior. But at that point the packaging script itself could easily do the same thing and add -DXZ_SANDBOX=no if -fsanitize= is found in CFLAGS.

Questions:

1. Does vcpkg have a builtin option to enable sanitizers, or do you enable it yourself by specifying extra CFLAGS?

2. If you are adding extra CFLAGS, couldn't you also pass -DXZ_SANDBOX=no to CMake? The variable XZ_SANDBOX is likely unique to the xz package, thus it should be safe to set globally if per-package setting is difficult.

3. If the sanitizer flags are added by vcpkg, could vcpkg's liblzma portfile check if sanitizers are enabled and then add -DXZ_SANDBOX=no?

bansan85:
1) I explicitly set `-fsanitize=thread` *_FLAGS in my triplet:

```
set(VCPKG_TARGET_ARCHITECTURE x64)
set(VCPKG_CRT_LINKAGE dynamic)
set(VCPKG_LIBRARY_LINKAGE dynamic)
set(VCPKG_CXX_FLAGS "-fsanitize=thread")
set(VCPKG_C_FLAGS "-fsanitize=thread")
set(VCPKG_CMAKE_CONFIGURE_OPTIONS -DCMAKE_CUDA_FLAGS="-fsanitize=thread")
set(VCPKG_LINKER_FLAGS "-fsanitize=thread")

set(VCPKG_CMAKE_SYSTEM_NAME Linux)
set(VCPKG_FIXUP_ELF_RPATH ON)
```

2) Yes, I can do something like:

```
if (PORT STREQUAL "liblzma")
  set(VCPKG_CMAKE_CONFIGURE_OPTIONS ${VCPKG_CMAKE_CONFIGURE_OPTIONS} -DXZ_SANDBOX=no)
endif()
```

But I don't like writing packet-specific things in the triplet. But custom triplet are not officially supported by vcpkg. So I try to reduce impact in port implementation.

3) Yes, I can modify the port to check if sanitize is enabled by reading VCPKG_C_FLAGS variable and updating XZ_SANDBOX variable.

There is a lot of solution. This is why I requested your opinion. To reduce maintenance in vcpkg project, patches are usually reported upstream. So if patching CMakeLists.txt is not the good solution, I will set XZ_SANDBOX.

Thanks

Larhzu:
Thanks for the answers! Seems that all choices have some downsides.

Since you are adding the -fsanitize flag yourself, your (2) doesn't sound too bad. However, I know very little about vcpkg, so I don't know if this solution is merely ugly or if it has bigger downsides.

The benefit of (2) is that it doesn't modify the port file, and thus there is no risk of new bugs in the port file.

In 2023, it was pondered if -fsanitize= with Landlock should be an error or warning. A warning would have worked for you if you only need the library (not the command line tools). I still think that an error is better in most situations, because a warning is easy to miss and then people would wonder why sanitizers fail with the command line tools.

So I won't change the upstream code. It was still good that you asked. It's useful for me to know what kind of use cases and needs there are at downstreams.

PR #210, opened 2026-02-23
CMake: Guard script configure_file() calls with ENABLE_SCRIPTS

rzikm:
The configure_file() calls for xzdiff, xzgrep, xzmore, and xzless were running unconditionally within the if(UNIX) block, even when ENABLE_SCRIPTS was OFF. This would cause a build failure if the src/scripts/*.in files were not present (we do that since we want to only build liblzma and link it to our application and want to minimize vendored code size)

Move the foreach loop inside the existing `if(ENABLE_SCRIPTS)` guard so that configure_file() is only called when scripts are actually being built and installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2405,20 +2405,20 @@ xzdiff, xzgrep, xzmore, xzless, and their symlinks" ON)
     set(xz "xz")
     set(POSIX_SHELL "${XZ_POSIX_SHELL}")
 
-    foreach(S xzdiff xzgrep xzmore xzless)
-        configure_file("src/scripts/${S}.in" "${S}"
-               @ONLY
-               NEWLINE_STYLE LF
-               FILE_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE
-                                GROUP_READ GROUP_EXECUTE
-                                WORLD_READ WORLD_EXECUTE)
+    if(ENABLE_SCRIPTS)
+        foreach(S xzdiff xzgrep xzmore xzless)
+            configure_file("src/scripts/${S}.in" "${S}"
+                   @ONLY
+                   NEWLINE_STYLE LF
+                   FILE_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE
+                                    GROUP_READ GROUP_EXECUTE
+                                    WORLD_READ WORLD_EXECUTE)
 
-        if(ENABLE_SCRIPTS)
             install(PROGRAMS "${CMAKE_CURRENT_BINARY_DIR}/${S}"
                     DESTINATION "${CMAKE_INSTALL_BINDIR}"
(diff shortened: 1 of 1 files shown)


Larhzu:
Thanks! I simplified it so that there is only one `if(ENABLE_SCRIPTS)` block. It's in the `misc` branch at the moment. I didn't merge it yet. Please check that the [modified commit](https://github.com/tukaani-project/xz/commit/45d519906ef3937d9058aa0a1cdd1df4799fa8b4), including the commit message, are OK.

I noticed that a few issues were found in [124003](https://github.com/dotnet/runtime/pull/124003) so I fixed those as well. The [comment](https://github.com/dotnet/runtime/pull/124003#discussion_r2799391812) about `_OPENBSD_SOURCE` I didn't understand though; it looks like a false alarm, but I mention it anyway in case I missed something.

The [file descriptor](https://github.com/dotnet/runtime/pull/124003#discussion_r2782313624) comment is a false positive. Opening with the opposite modes is intentional.

When sophisticated tools like static analyzers or LLMs are used to find and/or fix issues, mentioning it in the commit message might be good. However, `Co-authored-by` is a wrong trailer for this. A LLM isn't a legal entity, it cannot be a copyright holder, and it isn't responsible for the work.

When the work is very small in size (like this PR), it likely doesn't cross the copyrightability threshold. However, with bigger patches, the copyright status of LLM-generated works isn't clear to me, so I'm wary of accepting such patches for now.

rzikm:
Hey, thanks for taking time to go through the linked .NET runtime PR.

> I didn't merge it yet. Please check that the [modified commit](https://github.com/tukaani-project/xz/commit/45d519906ef3937d9058aa0a1cdd1df4799fa8b4), including the commit message, are OK.

LGTM!

> I noticed that a few issues were found in https://github.com/dotnet/runtime/pull/124003 so I fixed those as well. The https://github.com/dotnet/runtime/pull/124003#discussion_r2799391812 about _OPENBSD_SOURCE I didn't understand though; it looks like a false alarm, but I mention it anyway in case I missed something.

> 
> The https://github.com/dotnet/runtime/pull/124003#discussion_r2782313624 comment is a false positive. Opening with the opposite modes is intentional.

I didn't give too much attention to those LLM review comments because the intent of that PR is to vendor the source code with as little modifications as possible (to simplify future updates). My impression is that on giant PRs like those (+50k LOC) it might be more prone to hallucinations due to insufficient context window.

> When sophisticated tools like static analyzers or LLMs are used to find and/or fix issues, mentioning it in the commit message might be good. However, Co-authored-by is a wrong trailer for this. A LLM isn't a legal entity, it cannot be a copyright holder, and it isn't responsible for the work.

I will pass that to relevant feedback channel. Personally, I would be fine with omitting it and keeping only myself as the commit author, as I would make the same change manually and wouldn't raise a generated PR that wouldn't be up to my standard.

PR #214, opened 2026-03-04
Add linker flag check for output DEF file

VitorEAFeliciano:
let cmake find the correct export symbols with clang or msvc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,7 +92,10 @@ include(cmake/tuklib_cpucores.cmake)
 include(cmake/tuklib_physmem.cmake)
 include(cmake/tuklib_progname.cmake)
 include(cmake/tuklib_mbstr.cmake)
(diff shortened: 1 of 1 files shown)


Larhzu:
Problems:

  * `WINDOWS_EXPORT_ALL_SYMBOLS` puts all symbols to exports.def. A correct DEF file should only include those symbols that have been marked with `dllexport`.

  * If I remember correctly, one shouldn't use CMake's checks before the `project` command.

If I build with Clang/LLVM + mingw-w64, a DEF file is created just fine by the LLVM linker. So I assume that you wish to create a DEF file for liblzma.dll when building with clang-cl under MSVC or with MSVC itself.

I don't know a lot about developing for Windows. My understanding is that the DEF file was useful when liblzma couldn't be built with MSVC at all. One could build liblzma.dll with mingw-w64 and then create an import library for MSVC from the DEF file.

When building with clang-cl, the build creates an import library lzma.lib for liblzma.dll. Is that not enough?

Larhzu:
Closing due to no reply from the submitter. I don't even understand what exactly should be fixed.

PR #218, opened 2026-03-22
xz: Add overflow check to read_name() buffer doubling

kodareef5:
`read_name()` in `src/xz/main.c` doubles its filename buffer with `size *= 2` without checking for overflow. If `size` exceeds `SIZE_MAX / 2`, the multiplication wraps to 0 and `xrealloc(name, 0)` can free the buffer while subsequent code writes to it via `name[pos++] = c`.

This adds a check before the doubling that aborts with an error if the buffer has grown beyond `SIZE_MAX / 2`. This addresses the existing FIXME at line 61:

    // FIXME: Maybe we should have some kind of memory usage limit here

### DIFF
--- a/src/xz/main.c
+++ b/src/xz/main.c
@@ -134,6 +134,9 @@ read_name(const args_info *args)
 		// at least for one character to allow terminating the string
 		// with '\0'.
 		if (pos == size) {
+			if (size > SIZE_MAX / 2) {
+				message_fatal(_("Filename is too long"));
+			}
 			size *= 2;
 			name = xrealloc(name, size);
 		}


collinfunk:
This change seems fine to me, but in practice the existing code is mostly okay.

On typical platforms, where `PTRDIFF_MAX * 2` is equal to `SIZE_MAX - 1`, `xmalloc` would exit the program with an error because `malloc` cannot allocate more than `PTRDIFF_MAX` bytes.

Clashnewbme:
This change seems pretty nice, I dont see the reason to not accept it.

Larhzu:
I see that the @kodareef5 account was restored and this PR is visible again. Thanks for the report!

I think the overflow problem is somewhat theoretical, but it's good to check it anyway. Allocating `SIZE_MAX / 2 + 1` bytes might not succeed on many 32-bit platforms, but it's not good to rely on that. I committed a slightly different fix than the one proposed here.

@kodareef5, was this PR generated with AI/LLM? Certain details and wording make me suspect it, but I would like to be certain.

Note that this PR _doesn't_ address the `FIXME` about memory usage limit. xz supports memory usage limiting to prevent unexpectedly high memory usage, but those limits don't affect `--files` (and perhaps they don't need to either).

PR #221, opened 2026-05-10
Added better error codes for lzma_code()

Joosibaeri:
This change adds human-readable error messages for selected lzma_code() return values, making liblzma errors easier to understand and diagnose. A new public helper function, lzma_get_error_message(), lets applications retrieve the message associated with the most recent lzma_code() failure. The implementation includes support for common errors such as out-of-memory, data corruption, unsupported options, and programming errors, while returning NULL when no message is available. The new behavior is covered by tests and shown in the example programs, and the documentation has been updated accordingly.

As a first contribution, I’d appreciate any feedback on the code quality, naming, or implementation style to help align this change with the project’s standards.

### DIFF
--- a/TODO
+++ b/TODO
@@ -56,10 +56,6 @@ Missing features
     Support LZMA_FULL_FLUSH for lzma_stream_decoder() to stop at
     Block and Stream boundaries.
 
-    Error codes from lzma_code() aren't very specific. A more detailed
-    error message (string) could be provided too. It could be returned
-    by a new function or use a currently-reserved member of lzma_stream.
-
     Make it possible to adjust LZMA2 options in the middle of a Block
     so that the encoding speed vs. compression ratio can be optimized
     when the compressed data is streamed over network.
--- a/doc/examples/05_error_messages.c
+++ b/doc/examples/05_error_messages.c
@@ -1,0 +1,120 @@
+/*
+ * This example demonstrates how to retrieve detailed error messages
+ * from lzma_code() when an error occurs.
+ *
+ * SPDX-License-Identifier: 0BSD
+ */
+
+#include <stdio.h>
+#include <lzma.h>
+
+int
+main(void)
+{
+	printf("Example 1: Error message for programming error\n");
+	printf("-----------------------------------------------\n\n");
+
+	lzma_stream strm = LZMA_STREAM_INIT;
+	lzma_ret ret;
+
+	// Initialize a valid encoder
+	ret = lzma_easy_encoder(&strm, LZMA_PRESET_DEFAULT, LZMA_CHECK_CRC64);
+	if (ret != LZMA_OK) {
+		fprintf(stderr, "Failed to initialize encoder\n");
+		return 1;
+	}
+
+	// Try to call lzma_code with invalid arguments (NULL next_out with avail_out > 0)
+	// This causes a LZMA_PROG_ERROR
+	strm.next_out = NULL;
+	strm.avail_out = 100;
+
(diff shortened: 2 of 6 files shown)


Joosibaeri:
Sorry, I dont want a review from Copilot but I forgot to turn it of in the settings...

Larhzu:
I see you closed the PR but didn't mention why, so here's some feedback.

It's good to have a description in the PR, but even more important is to have it in the commit message(s). The commit messages are much more likely to remain available in the future than the PR discussion.

The changes should be split into multiple commits when it's easy. For example:

1. Add the feature.
2. Add the tests.
3. Add the examples.
4. Update TODO.

In this project, `SPDX-License-Identifier` should be on the first line of the file when possible.

Line length should stay below 80 chars whenever it's practical enough.

This PR maps `lzma_ret` values to strings in a static manner. For static mapping, it would be enough to have a function that takes a `lzma_ret` argument and returns `const char *` to the matching string. It would be much simpler than the method in this PR. However, the method in this PR can be extended to support more specific messages in specific situation, so that is an upside.

If a pointer to the message is stored directly in `lzma_stream`, there's no need to have a separate function to fetch the message; apps can read the message directly from `lzma_stream`. One could rename the member from `reserved_ptr1` to something else like `errmsg` or `error_msg`.

To provide more detailed error messages, the message would need to be set deeper in the call stack. For example, in src/liblzma/common/stream_decoder.c, if an error occurs when verifying the Index field in `SEQ_INDEX`, then the error message could tell that. Compare to XZ for Java where the exceptions include a string that tell a little bit more about the error ([example][1]).

In liblzma, the error message pointer would need to be accessible in places that cannot access `lzma_stream` directly. One could either pass the address of `lzma_stream.error_msg` when initializing a coder or by adding an extra argument to [lzma_code_function][2]. There are pros and cons to both. [The first method][3] is used with `lzma_file_info_decoder` where `&strm->seek_pos` is [stored in the internal state][4].

It could be nicer if the error message is never set to NULL. That way the string is simpler and safer to use.

The TODO entry is old. I have mixed feelings about it. More detailed message would have some value when debugging what's wrong with a file, but I'm not sure if it is worth the hassle to add them now. The much simpler static mapping to messages wasn't done mostly because with some error codes the best message depends on which API function returned it. But on the other hand, xz itself [doesn't do anything better][5] either.

> Sorry, I dont want a review from Copilot but I forgot to turn it of in the settings...

:-) Seems that all of its comments point out real issues in the PR though.

[1]: https://github.com/tukaani-project/xz-java/blob/492b6ea034b24427365d760721f14c971f9d0d01/src/org/tukaani/xz/SingleXZInputStream.java#L452
[2]: https://github.com/tukaani-project/xz/blob/1aab7e9cab1293c2b8a481ed1beac94da21ef04f/src/liblzma/common/common.h#L185
[3]: https://github.com/tukaani-project/xz/blob/1aab7e9cab1293c2b8a481ed1beac94da21ef04f/src/liblzma/common/file_info.c#L838
[4]: https://github.com/tukaani-project/xz/blob/1aab7e9cab1293c2b8a481ed1beac94da21ef04f/src/liblzma/common/file_info.c#L820
[5]: https://github.com/tukaani-project/xz/blob/1aab7e9cab1293c2b8a481ed1beac94da21ef04f/src/xz/message.c#L798

Joosibaeri:
Thanks for the review, I got similar feedback also on one of my own repositories. Perhaps one could update the todos for future (specifically for new maintainers like me).

PR #222, opened 2026-05-27
xz: Mask the multiplier suffix in str_to_uint64 error message

rootvector2:
The "Invalid multiplier suffix" error in str_to_uint64 prints the user-supplied suffix directly, while the other error path in the same function masks it. A suffix containing control bytes therefore reaches the terminal unescaped. Commit 7e7bb6c masked the sibling message a few lines up but missed this one, so route it through tuklib_mask_nonprint too.

### DIFF
--- a/src/xz/util.c
+++ b/src/xz/util.c
@@ -154,7 +154,7 @@ str_to_uint64(const char *name, const char *value, uint64_t min, uint64_t max)
 
 		if (multiplier == 0) {
 			message(V_ERROR, _("%s: Invalid multiplier suffix"),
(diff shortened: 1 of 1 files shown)


Clashnewbme:
This is a great fix, and helps the "Invalid multiplier suffix" error, nice.

Larhzu:
Thanks!

When I made the commit d22f96921fd2f94d842f3cc2e5f729cb3cca5122, I thought that it's mostly filenames that matter, but I agree it's better to mask the other strings too.

PR #223, opened 2026-05-28
xz: Copy --files/--files0 path so it outlives XZ_OPT/XZ_DEFAULTS parsing

rootvector2:
Noticed args->files_name is set straight from optarg, but when the option comes from XZ_OPT or XZ_DEFAULTS, parse_environment() frees the env buffer optarg points into. Later reads through args->files_name (e.g. the 'Unexpected end of input' / null-character paths in read_name()) hit freed memory.

Trigger: `printf foo.txt > files.lst; XZ_OPT=--files=files.lst xz` prints `xz: ?: Unexpected end of input` instead of `xz: files.lst: ...` because tuklib_mask_nonprint walks the now-overwritten bytes. xstrdup the path so args->files_name owns it.

### DIFF
--- a/src/xz/args.c
+++ b/src/xz/args.c
@@ -641,15 +641,20 @@ parse_real(args_info *args, int argc, char **argv)
 				args->files_name = stdin_filename;
 				args->files_file = stdin;
 			} else {
-				args->files_name = optarg;
-				args->files_file = fopen(optarg,
+				// optarg may point into a buffer that doesn't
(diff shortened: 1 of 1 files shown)


Larhzu:
Good catch, thanks! I edited the commit message and the comment in the code.

There didn't seem to be other similar bugs. For example, `suffix_set(optarg)` makes its own copy of the string.

PR #224, opened 2026-05-29
fuzz: add index decoder, stream flags, and block header fuzz harnesses

XananasX7:
## Summary

This PR adds three new fuzz harnesses to `tests/ossfuzz/` that cover liblzma APIs not previously fuzzed by any existing harness.

### `fuzz_index_decoder.c`
Fuzzes the `.xz` **Index decoder** using both:
- `lzma_index_buffer_decode()` — single-call interface
- `lzma_index_decoder()` — streaming interface via `lzma_code()`

The Index is a critical structure at the end of every `.xz` stream that describes all Block boundaries. Parsing bugs here could affect random-access reading and integrity checking. The harness also walks decoded indexes with `lzma_index_iter` to exercise the iterator code paths.

### `fuzz_stream_flags.c`
Fuzzes **Stream Header / Footer flags parsing**:
- `lzma_stream_header_decode()` + `lzma_stream_header_encode()` (round-trip)
- `lzma_stream_footer_decode()` + `lzma_stream_footer_encode()` (round-trip)
- `lzma_stream_flags_compare()` when two full-size buffers are available

Stream flag fields encode the check type and other per-stream metadata; errors here affect every `.xz` file processed by the library.

### `fuzz_block_header.c`
Fuzzes **Block Header decoding** via `lzma_block_header_decode()`. Block headers carry the filter chain used to compress each Block. Parsing bugs could trigger memory-safety issues when constructing filter option structures or walking variable-length filter lists. The harness correctly initialises `lzma_block`, handles the header-size byte, and frees filter options with `lzma_filters_free()` on success.

## What is NOT duplicated

The existing harnesses cover stream-level decode/encode (`fuzz_decode_stream`, `fuzz_decode_stream_mt`, `fuzz_encode_stream`) and the legacy `.lzma` format (`fuzz_decode_alone`). None of those exercise the Index decoder, Stream flags, or Block header APIs directly.

## Harness quality

- `LLVMFuzzerTestOneInput()` signature, always returns 0
- Aborts only on allocation failures (not parse errors)
- Minimum size guards to skip uninteresting short inputs
- `.options` files in `tests/ossfuzz/config/` with `max_len = 65536`
- No Makefile changes needed — the existing `$(wildcard *.c)` rule picks up all new targets automatically

## Context

These harnesses were developed for the [Google OSS-Fuzz](https://github.com/google/oss-fuzz) integration to increase coverage of security-critical liblzma parsing paths.

### DIFF
--- a/tests/ossfuzz/config/fuzz_block_header.options
+++ b/tests/ossfuzz/config/fuzz_block_header.options
@@ -1,0 +1,4 @@
+# SPDX-License-Identifier: 0BSD
+
+[libfuzzer]
+max_len = 65536
--- a/tests/ossfuzz/config/fuzz_index_decoder.options
+++ b/tests/ossfuzz/config/fuzz_index_decoder.options
@@ -1,0 +1,4 @@
+# SPDX-License-Identifier: 0BSD
+
+[libfuzzer]
+max_len = 65536
--- a/tests/ossfuzz/config/fuzz_stream_flags.options
+++ b/tests/ossfuzz/config/fuzz_stream_flags.options
@@ -1,0 +1,4 @@
+# SPDX-License-Identifier: 0BSD
+
+[libfuzzer]
+max_len = 65536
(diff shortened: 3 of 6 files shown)


Larhzu:
Index decoder fuzzing was already added. The others should already be fuzzed indirectly via fuzz_\*_stream*.c files.

PR #225, opened 2026-05-29
xz: mask non-printable chars in the invalid multiplier suffix message

rootvector2:
str_to_uint64() masks non-printable bytes in the "not a non-negative decimal integer" message (commit 7e7bb6c) but the "Invalid multiplier suffix" message a few lines down still prints the raw suffix. A leading digit routes parsing into that branch, so `--block-size=5$'\e[2J'` (or the same via XZ_OPT) sends control/escape bytes straight to the terminal. Mask it the same way.

### DIFF
--- a/src/xz/util.c
+++ b/src/xz/util.c
@@ -154,7 +154,7 @@ str_to_uint64(const char *name, const char *value, uint64_t min, uint64_t max)
 
 		if (multiplier == 0) {
 			message(V_ERROR, _("%s: Invalid multiplier suffix"),
-					value - 1);
+					tuklib_mask_nonprint(value - 1));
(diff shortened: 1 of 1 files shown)


Larhzu:
This is a duplicate of #222 by the same GitHub user but different message and commit message. I'm curious, is it an honest mistake by a busy human (it's fine), or is this a case of a cat being away and now AI bots will play?

rootvector2:
No it's a human. And sorry i forgot about the previous pr..

rootvector2:
And honestly I want to tell you I am busy with others works..

Larhzu:
Alright, no problem then. :-)

PR #226, opened 2026-05-31
file_info: preserve unrepresentable memusage and unlimited memlimit semantics

jmestwa-coder:
## Summary

Fix two related correctness issues in `file_info_decoder_memconfig()` when handling very large Index memory requirements.

### Changes

* Prevent overflow when combining `combined_index_memusage` and `this_index_memusage`.
* Return `UINT64_MAX` when the combined memory requirement cannot be represented.
* Preserve the documented `UINT64_MAX` ("disable the limiter") memlimit semantics when forwarding memory limits to the nested Index decoder.

### Problem

A crafted concatenated `.xz` file could cause the combined memory usage calculation to overflow, making `lzma_memusage()` report a wrapped value instead of an unrepresentable requirement.

After correcting the reported memory usage, calling:

```c
lzma_memlimit_set(&strm, UINT64_MAX);
```

could still fail because the unlimited memlimit sentinel was converted into a finite value before being passed to the nested decoder.

### Result

* `lzma_memusage()` now returns `UINT64_MAX` for unrepresentable combined memory requirements.
* `lzma_memlimit_set(UINT64_MAX)` correctly preserves unlimited memlimit semantics.
* Decoding proceeds past the memory-limit check and reports the malformed stream with the expected error.

### DIFF
--- a/src/liblzma/common/file_info.c
+++ b/src/liblzma/common/file_info.c
@@ -722,7 +722,11 @@ file_info_decoder_memconfig(void *coder_ptr, uint64_t *memusage,
 	// acceptable as lzma_memusage() has to return non-zero on success
 	// and even with an empty .xz file we will end up with a lzma_index
 	// that takes some memory.
-	*memusage = combined_index_memusage + this_index_memusage;
+	if (UINT64_MAX - combined_index_memusage < this_index_memusage)
+		*memusage = UINT64_MAX;
+	else
+		*memusage = combined_index_memusage + this_index_memusage;
+
 	if (*memusage == 0)
 		*memusage = lzma_index_memusage(1, 0);
 
@@ -737,8 +741,11 @@ file_info_decoder_memconfig(void *coder_ptr, uint64_t *memusage,
 		// its new memory usage limit.
 		if (coder->this_index == NULL
 				&& coder->sequence == SEQ_INDEX_DECODE) {
-			const uint64_t idec_new_memlimit = new_memlimit
-					- combined_index_memusage;
+			const uint64_t idec_new_memlimit
+					= new_memlimit == UINT64_MAX
+						? UINT64_MAX
+						: new_memlimit
+							- combined_index_memusage;
 
 			assert(this_index_memusage > 0);
 			assert(idec_new_memlimit > 0);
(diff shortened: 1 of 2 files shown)


Larhzu:
Thanks for the report!

> Prevent overflow when combining combined_index_memusage and this_index_memusage.

The main issue was that index_decoder.c could return crazy high memusage when Number of Records is far larger than what is possible in a valid .xz file. file_info.c didn't expect such values.

I fixed the root cause (index_decoder.c) and also added an overflow check as an extra caution to file_info.c. The max value from index_decoder.c is now less than 130 GiB, which obviously is still a huge value, but it's tiny compared to `UINT64_MAX`.

> Return UINT64_MAX when the combined memory requirement cannot be represented.

I made `file_info_decoder_memconfig` fail with `LZMA_PROG_ERROR` if the overflow happens (even though it shouldn't be possible now).

> Preserve the documented UINT64_MAX ("disable the limiter") memlimit semantics when forwarding memory limits to the nested Index decoder.

This would be inconsistent with the `SEQ_INDEX_INIT` case in `file_info_decode`. After the index_decoder.c fix it also makes no practical difference because memory allocation will fail before getting anywhere near `UINT64_MAX`. Even if allocation didn't fail, it might still be impossible to create a .xz file where the overflow could be triggered. I didn't calculate it exactly but the required file size might exceed 8 EiB, that is, the maximum size of a valid .xz file and signed 64-bit `off_t`. Anyway, now with the overflow check the behavior will be acceptable enough in this corner case.

My changes broke the test case added in this PR. This was expected. I'm not sure if some other test could be added now. The new `INDEX_RECORDS_MAX` check is a very simple addition. The memory usage is untestable apart from checking that the value won't be insanely high.

jmestwa-coder:
Thanks for investigating and for the detailed explanation.

That makes sense. I was focused on the overflow in `file_info.c`, but your explanation of the root cause in `index_decoder.c` and the new `INDEX_RECORDS_MAX` check clarifies why those extreme memusage values shouldn't be reachable in the first place.

I also understand your point about the `UINT64_MAX` memlimit handling being inconsistent with the `SEQ_INDEX_INIT` path and not providing practical benefit after the root-cause fix.

PR #227, opened 2026-06-03
ci: pin GitHub Actions to full commit SHAs

XananasX7:
Several workflow files reference GitHub Actions via mutable version tags (e.g. `@v4`, `@v3`) instead of full commit SHAs. This is a supply chain risk — a compromised action repository could silently alter CI behavior.

This PR pins each action to the exact commit SHA corresponding to the version tag, making the dependency immutable and auditable.

Recommended by [GitHub security hardening guide](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) and [OpenSSF Scorecard](https://securityscorecards.dev/).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 20
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
       ########################
       # Install Dependencies #
@@ -164,7 +164,7 @@ jobs:
         run: ./build-aux/ci_build.bash -b ${{ matrix.build_system }} -d small -p test -n small
 
       # Attempt to upload the test logs as artifacts if any step has failed
-      - uses: actions/upload-artifact@v7
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         if: ${{ failure() }}
         with:
           name: ${{ matrix.os }} ${{ matrix.build_system }} Test Logs
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -28,7 +28,7 @@ jobs:
     steps:
     - name: Build Fuzzers (${{ matrix.sanitizer }})
       id: build
-      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@10d372d86bc3ca5d662790345c2f4e146367e4c4 # master
       with:
         oss-fuzz-project-name: 'xz'
         # The language must match the one in project.yaml in OSS-Fuzz:
@@ -38,7 +38,7 @@ jobs:
         sanitizer: ${{ matrix.sanitizer }}
 
     - name: Run Fuzzers (${{ matrix.sanitizer }})
-      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@10d372d86bc3ca5d662790345c2f4e146367e4c4 # master
       with:
         oss-fuzz-project-name: 'xz'
         language: c++
@@ -48,7 +48,7 @@ jobs:
         report-ooms: true
 
     - name: Upload Crash
-      uses: actions/upload-artifact@v7
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
       if: failure() && steps.build.outcome == 'success'
       with:
         name: ${{ matrix.sanitizer }}-artifacts
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -12,7 +12,7 @@ jobs:
   coverity:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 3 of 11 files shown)


thesamesam:
There's no changes in the diff.

XananasX7:
Apologies for the empty diff — the branch had an identity-fix commit on top with no actual file changes underneath it. Fixed now: all 11 workflow files are updated with full commit SHAs for `actions/checkout@v6`, `actions/upload-artifact@v7`, and `google/oss-fuzz@master`, each annotated with the version tag in a comment for readability.

XananasX7:
Friendly ping — happy to address any review comments. Thank you!

Larhzu:
Sorry, I have to prioritize. There are many things waiting for reply or other action.

In the commit 35f8649f0834 it was decided that GitHub's own actions are considered trusted and thus won't be pinned to commit ID. Is there some reason, even a minor one, why this should be reconsidered?

The cifuzz actions are from Google, so the above rule doesn't cover it.

The workflow YAML files have `permissions: {}` (except coverity.yml, which likely should be fixed). The organization (and thus all projects in it) have *Workflow permissions* set to *Read repository contents and packages permissions*, and thus *Allow GitHub Actions to create and approve pull requests* is forced off too. Do these prevent malicious actions from modifying the repositories and from reading secrets that haven't been listed in the YAML file? If yes, does the paranoia level need to be so high, that is, perhaps not everything needs to be pinned?

As an extra, I guess the `on: pull_request:` part from the YAML files could be removed because I don't run CI from PRs anyway.

@Alb3e3: You created a related PR #230, so I'm tagging you in case you happen to have something to comment here.

XananasX7:
Thanks for the detailed response @Larhzu — and no worries about the delay, I appreciate you taking the time.\n\nYou're right about the project's existing policy: GitHub-operated actions (`actions/checkout`, `actions/upload-artifact`) are considered trusted and won't be pinned. I should have checked commit 35f8649f0834 before submitting.\n\nThe cifuzz actions from Google are a different story though — they're not GitHub-operated, so the PR's argument applies there.\n\nI can rework this to only pin the `google/cifuzz` and `google/oss-fuzz` action references and drop all the `actions/*` pins entirely. On `coverity.yml` — agreed, if the other files have `permissions: {}` it would be consistent to add it there too; happy to include that fix as well.\n\nShould I push an updated commit with that scope? Let me know.

Alb3e3:
Thanks for tagging me. My PR #230 is intentionally scoped only to the `coverity.yml` permissions gap mentioned here, and does not try to pin the GitHub-owned actions.

My understanding is that the organization-level read-only workflow permission setting and explicit workflow permissions prevent the workflow token from getting repository write access, and secrets are not enumerable unless the workflow explicitly passes them into a job or step. That makes broad pinning of `actions/*` less compelling if the project policy is to treat GitHub-owned actions as trusted.

`coverity.yml` is a narrower case because it was the exception without an explicit permissions block and it uses Coverity credentials in its workflow. PR #230 just makes the token scope explicit with `contents: read` while preserving the existing Coverity scan behavior.

thesamesam:
@eli-schwartz makes a similar argument at https://github.com/rockdaboot/libpsl/pull/266#pullrequestreview-4493707892.

eli-schwartz:
I have significant doubts that this PR was produced outside of an LLM, for a couple reasons but the newline habits of the recent reply feel like a solid giveaway.

In the linked PR, there was also an assertion that the PR should be acknowledged to qualify for some Google bug bounty which kind of gave me a bad feeling about the whole process. I also checked submitter profile history and I'm automatically biased against any user account that files PRs against hundreds of projects a day for the same change...

Larhzu:
There are only so many ways to pin actions to commit IDs, so humans and LLMs may produce identical diffs in this case. I didn't think much about @XananasX7's initial message, but the [second message](https://github.com/tukaani-project/xz/pull/227#issuecomment-4610143833) had clear LLM vibes. The [fourth message](https://github.com/tukaani-project/xz/pull/227#issuecomment-4705280466) with `\n` and a bunch of other clues removed all doubts.

Anyway, the diff only works as an example because the maintainer *must* verify that the commit IDs are correct. It can be easiest to do the commit from scratch and compare against the PR afterwards. (A malicious PR could attempt to pin the actions to commit IDs of a malicious fork.)

With changes like this, the value of the PR is in the discussion and commit message, not in the diff itself.

@Alb3e3 wrote:
> My understanding is that the organization-level read-only workflow permission setting and explicit workflow permissions prevent the workflow token from getting repository write access, and secrets are not enumerable unless the workflow explicitly passes them into a job or step. That makes broad pinning of `actions/*` less compelling if the project policy is to treat GitHub-owned actions as trusted.

Thanks! I don't see strong reasons to limit this to `actions/*`. If a workflow has no secrets other than `GITHUB_TOKEN`, and that only has `contents: read` permission, then pinning any action doesn't seem super important.

If an action was compromised, it might use the CPU time and Internet connection of the runner on GitHub until workflow's timeout is reached (10-30 minutes). For example, a compromised action could run a cryptocoin miner or DDoS botnet node for a few minutes. That is bad, but at least it should have no access to project's code, secrets, or other things.

I mentioned this to @eli-schwartz and @thesamesam on IRC. I didn't hear opposing arguments.

If the actions aren't pinned, then the workflows will use the latest actions instead of several months old actions, and the hassle of updating the commit IDs is avoided. The benefits of not pinning seem higher than the risks *when the workflow is read only and uses no secrets* (other than `GITHUB_TOKEN`).

I removed the pins from all workflows except coverity.yml which uses secrets, see the commit 0def41200695.

Thanks to everyone who helped!

PR #230, opened 2026-06-11
Harden Coverity workflow permissions

Alb3e3:
This tightens the GitHub Actions Coverity workflow permissions without changing the scan behavior.

The workflow only needs repository read access for checkout and build execution, so this sets the default `GITHUB_TOKEN` permission to `contents: read`.

I verified the workflow YAML still parses, `git diff --check` passes, and `zizmor` no longer reports the previous excessive-permissions finding. The remaining finding is an action pinning warning, which I left out of this focused change.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -8,6 +8,9 @@ on:
   push:
     branches: [coverity_scan]
 
+permissions:
+  contents: read
+
 jobs:
   coverity:
     runs-on: ubuntu-latest


Larhzu:
Thanks!

I think it should work even without `contents: read` because the other workflows work that way too, so I committed that. The checkout action is [documented](https://github.com/actions/checkout#recommended-permissions) to need it but with a public repository even that doesn't seem mandatory.

PR #231, opened 2026-06-22
Tests: Add tests for single-call .xz Stream encoding

androvonx95:
Title: Tests: Add tests for single-call .xz Stream encoding

There's already a test for the single-call decoder (test_stream_buffer_decode),
but the encoder side wasn't tested, so this adds one for it.

It covers lzma_stream_buffer_bound(), lzma_easy_buffer_encode(), and
lzma_stream_buffer_encode(). For the encoders it compresses some data and then
decodes it again to make sure the output round-trips back to the original. It
also checks the error cases: too-small output buffer, an invalid check or NULL
filters, and an unsupported check type.

The test is added to both the Autotools and CMake builds and uses the usual
HAVE_ENCODERS/HAVE_DECODERS guards so it skips cleanly when those are disabled.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -40,6 +40,7 @@ check_PROGRAMS = \
 	test_check \
 	test_hardware \
 	test_stream_buffer_decode \
+	test_stream_buffer_encode \
 	test_stream_flags \
 	test_filter_flags \
 	test_filter_str \
@@ -55,6 +56,7 @@ TESTS = \
 	test_check \
 	test_hardware \
 	test_stream_buffer_decode \
+	test_stream_buffer_encode \
 	test_stream_flags \
 	test_filter_flags \
 	test_filter_str \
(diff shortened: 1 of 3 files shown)


Larhzu:
Thanks! In this particular case I felt it was quicker for me to rewrite the tests instead of editing your patch because there were several details I wanted to do slightly differently. I put your GitHub username into `Thanks-to`.

I appreciate that you mentioned LLM usage. However, `Co-authored-by` is a wrong trailer to document LLM usage. A LLM isn't a legal entity, it cannot be a copyright holder, and it isn't responsible for the work. With tiny commits (under 20-30 lines), I don't know if adds any value to indicate the LLM tooling; such commits a human would likely have done very similarly anyway. With non-tiny patches like this PR, the copyright status of LLM-generated works isn't clear to me, so I'm wary of accepting such patches for now. Note that even if this PR wasn't generated with LLM, I would have done the rewrite still; the LLM usage wasn't the main reason.

PR #235, opened 2026-07-11
xz: Respect cgroup v2 memory limits on Linux

### LINKED ISSUE #234: Respect cgroup v2 memory limits on Linux
Follow-up to #195.

I ran into the same general problem from a slightly different angle: xz was running inside a memory-limited Linux container, but it saw the
host's physical RAM instead of the container's limit. It then picked a host-sized automatic memory limit and could try to use far more memory
than the container had available, resulting in allocation failures or an OOM kill for xz or other processes in that container.

The rlimit-based fix for #195 doesn't catch this setup. Containers commonly use cgroup v2 `memory.max` without setting `RLIMIT_DATA` or `RLIMIT_AS`.

This should be addressed because Linux containers are quite common.

## Example

On a host with about 63 GiB of RAM, run xz in a transient 256 MiB cgroup:

```sh
sudo systemd-run --wait --pipe --collect -p MemoryMax=256M \
    xz --robot --info-memory
```

Current xz reports host-sized values:

```text
67317026816    0    0    16829256704    16829256704    8
```

The first field is about 63 GiB of physical RAM and the automatic multithreaded memory limit is about 15.7 GiB, even though the process is limited to 256 MiB.

## Expected behavior

I think that when a finite cgroup v2 `memory.max` is lower than physical RAM, xz should use it as the RAM basis for:

- the automatic multithreaded memory limit;
- percentage-based `--memlimit-*` options.

Physical RAM can remain the fallback when no tighter limit is known and can still be reported by `--info-memory`. The existing rlimit and
32-bit address-space caps should still lower the automatic limit after the cgroup limit has been applied.

Explicit byte limits such as `--memlimit-compress=4GiB` should remain user-controlled.

I have a PR prepared to address this.

danielrobbins:
This fixes issues with xz exhausting memory in containers, causing it and other processes (due to OOM killer) to fail.

When xz runs in a memory-limited Linux container, sysinfo() and sysconf() may report the RAM of the host instead of the container's memory limit. As a result, xz can choose too many threads or a percentage-based memory limit that exceeds the container's allocation, causing allocation failures or an OOM kill.

This fix determines the effective cgroup v2 memory limit and use it as an upper bound for the RAM available to xz. Physical RAM remains the fallback and is still reported by --info-memory. Existing rlimit and 32-bit address-space caps continue to lower the automatic default afterward.

This also makes percentage-based --memlimit options use the same container-aware RAM basis. Add a Linux integration test using a 256 MiB transient systemd cgroup.

Fixes #234.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -614,6 +614,8 @@ set(LIBLZMA_API_HEADERS
 add_library(liblzma
     src/common/mythread.h
     src/common/sysdefs.h
+    src/common/tuklib_cgroup.c
+    src/common/tuklib_cgroup.h
     src/common/tuklib_common.h
     src/common/tuklib_config.h
     src/common/tuklib_integer.h
--- a/INSTALL
+++ b/INSTALL
@@ -555,7 +555,7 @@ XZ Utils Installation
                 detect the amount of physical memory on the system. This
                 information is used by the options --memlimit-compress,
                 --memlimit-decompress, and --memlimit when setting the
-                limit to a percentage of total RAM.
+                limit to a percentage of RAM available to xz.
 
                 On some systems, there is no code to detect the amount of
                 RAM though. Using --enable-assume-ram one can set how much
@@ -973,4 +973,3 @@ XZ Utils Installation
     about liblzma.so (or a similarly named file) being missing, try
     running "ldconfig" to update the run-time linker cache (if your
     operating system has such a command).
-
--- a/src/common/tuklib_cgroup.c
+++ b/src/common/tuklib_cgroup.c
@@ -1,0 +1,193 @@
+// SPDX-License-Identifier: 0BSD
+
+///////////////////////////////////////////////////////////////////////////////
+//
+/// \file       tuklib_cgroup.c
+/// \brief      Read Linux cgroup v2 limits
+//
+//  Author:     Lasse Collin
+//
+///////////////////////////////////////////////////////////////////////////////
+
+#include "tuklib_cgroup.h"
+
+#ifndef __linux__
+
+// Prevent an empty translation unit.
+typedef int dummy;
+
+#else
+
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <stdlib.h>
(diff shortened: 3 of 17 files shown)


danielrobbins:
@Larhzu I rebased based on your committed fixes, and added some correctness fixes. Descriptions are in the commit messages. This will make things more robust and cleans up some issues.

Larhzu:
Thanks! I updated the [cgroup branch](https://github.com/tukaani-project/xz/compare/master...cgroup). I added the `systemd-run true` test unconditionally. It's one more log entry, but if log entries are OK, then one more is OK too. I made other minor tweaks to the test too.

I'm not sure about the test without controller files. I have manually tested the code by looking at `strace` output, and the `fd != -1` check is unlikely to be removed from `get_cgroup_limit` by accident. I know the `trap` should ensure that the directory is always removed, so the test shouldn't leave anything behind, but I'm still not sure if the extra complexity is worth it considering what it tests.

It was pointed out to me that this kind of change might require [extra changes](https://github.com/SELinuxProject/refpolicy/commit/1e55618a24da234d8d3e629d9ec942e6026f7d94) if SELinux is used. Since this change affects liblzma, it can affect apps that call `lzma_physmem` or `lzma_cputhreads`. While it's only a small minor thing, it's still an extra thing for some distros to handle.

I'm leaning towards not including this in 5.8.4, which I try to release this month. It will have a bunch of other fixes, and I'd hope the release to be a simple smooth update for distros.

collinfunk:
> It was pointed out to me that this kind of change might require [extra changes](https://github.com/SELinuxProject/refpolicy/commit/1e55618a24da234d8d3e629d9ec942e6026f7d94) if SELinux is used. Since this change affects liblzma, it can affect apps that call lzma_physmem or lzma_cputhreads. While it's only a small minor thing, it's still an extra thing for some distros to handle.

FYI, @pixelb, not sure if you were aware of this, but our changes to make `nproc`, etc. cgroup aware required some SELinux changes. Just something for us to consider for future changes. Perhaps they would appreciate a heads up.

PR #236, opened 2026-07-22
  Add RISC-V Vector (RVV) optimized CRC32 implementation

dingcyrus:
When the compiler defines __riscv_vector (e.g. -march=rv64gcv),
  both the generic slice-by-eight and the RVV-accelerated CRC32 are
  built, with runtime selection via the existing crc32_func dispatch
  mechanism.

    - crc_common.h: Add CRC32_RISCV_VECTOR configuration, enabled when __riscv_vector is defined. Both CRC32_GENERIC and CRC32_ARCH_OPTIMIZED are set to support runtime dispatch.

    - crc32_rvv.h: New file providing crc32_arch_optimized() and is_arch_extension_supported(), following the same pattern as crc32_arm64.h and crc32_loongarch.h. Includes RVV helper macros inline.

    - crc32_fast.c: Include crc32_rvv.h when CRC32_RISCV_VECTOR is defined.

    - sysdefs.h: Guard the _Bool typedef with C99 version and C++ checks to avoid redefining _Bool where it is a built-in keyword.

    - Makefile.inc: Add crc32_rvv.h to liblzma_la_SOURCES.

### DIFF
--- a/src/common/sysdefs.h
+++ b/src/common/sysdefs.h
@@ -159,8 +159,12 @@
 #ifdef HAVE_STDBOOL_H
 #	include <stdbool.h>
 #else
-#	if ! HAVE__BOOL
+#	if !defined(__STDC_VERSION__) || __STDC_VERSION__ < 199901L
+#	ifndef __cplusplus
+#		if ! HAVE__BOOL
 typedef unsigned char _Bool;
+#		endif
+#	endif
 #	endif
 #	define bool _Bool
 #	define false 0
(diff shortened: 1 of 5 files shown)


Larhzu:
The sysdefs.h change is unrelated and should be in a separated commit. I don't know why it would be needed though because `HAVE__BOOL` should be defined when needed. I plan to cleanup legacy things like this once I bump version to 5.9.0alpha.

Did you test how much the CRC32 performance improves with this change?

Larhzu:
I'm closing this because no reply from the submitter and the patch obviously doesn't do anything useful in its current state.

PR #238, opened 2026-07-28
liblzma: Preserve check statistics in lzma_index_dup()

carrerasdarren-cell:
## Summary

- copy the cached integrity Check mask when duplicating an `lzma_index`
- add regression coverage for a three-Stream Index using CRC32, CRC64, and SHA-256

## Problem

`index_dup_stream()` preserves each Stream's flags, but `lzma_index_dup()` left the aggregate `checks` cache at its zero-initialized value. `lzma_index_checks()` therefore reported only the final Stream's Check type for a duplicated multi-Stream Index.

Before the fix, the regression test reports `1024` (SHA-256 only) instead of `1042` (CRC32, CRC64, and SHA-256). Copying `src->checks` preserves the same statistics as the source Index.

## Testing

- `ctest --test-dir /tmp/xz-build-normal-20260728-a -R '^test_index$' --output-on-failure`
- `ctest --test-dir /tmp/xz-build-normal-20260728-a --output-on-failure -j2` (20/20 passed)

### DIFF
--- a/src/liblzma/common/index.c
+++ b/src/liblzma/common/index.c
@@ -974,6 +974,7 @@ lzma_index_dup(const lzma_index *src, const lzma_allocator *allocator)
 	dest->total_size = src->total_size;
 	dest->record_count = src->record_count;
 	dest->index_list_size = src->index_list_size;
+	dest->checks = src->checks;
 
 	// Copy the Streams and the groups in them.
 	const index_stream *srcstream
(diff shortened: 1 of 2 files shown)


Larhzu:
Thanks! Did you find it because you needed the feature or by some other means? It wouldn't surprise me if no one had actually needed this combination of function calls, and thus the bug wasn't found before.

carrerasdarren-cell:
Thanks for applying it. I found this during a focused source/API review of `lzma_index_dup()`: the per-Stream flags were copied, but the aggregate `checks` cache was not. I then wrote the multi-Stream regression to confirm the public API mismatch; I had not encountered it through an application dependency.

Larhzu:
OK. :-) Thanks!

PR #239, opened 2026-08-12
Fix failing test cases on Windows

BraveChicken1:
This PR fixes failing test cases when running tests on Windows.
The issue only occurs when building xz using cmake, with option `-DBUILD_SHARED_LIBS=ON`. The test binaries cannot find the `liblzma.dll` file, resulting in a fail of the test cases.

To fix this issue, this PR introduces a copy command, which copies the `libzma.dll` file to the same directory as the test binaries.

### DIFF
--- a/tests/tests.cmake
+++ b/tests/tests.cmake
@@ -93,6 +93,16 @@ if(BUILD_TESTING)
             RUNTIME_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/tests_bin"
         )
 
+        # Copy liblzma.dll to the test directory on Windows if building shared
+        # libs because it has to be in the same directory as the test binaries.
+        if(WIN32 AND BUILD_SHARED_LIBS)
+            add_custom_command(TARGET "${TEST}" POST_BUILD
+                COMMAND ${CMAKE_COMMAND} -E copy_if_different
+                    $<TARGET_FILE:liblzma>
+                    $<TARGET_FILE_DIR:${TEST}>
+            )
+        endif()
+
         add_test(NAME "${TEST}"
                  COMMAND "${CMAKE_CURRENT_BINARY_DIR}/tests_bin/${TEST}"
         )


Larhzu:
With mingw-w64 + GCC + CMake + Ninja under MSYS2, it works for me. What toolchain are you using?

Looking at the [DLL search order](https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order), if the current directory is the build directory, then the test_*.exe files should find liblzma.dll from the current directory ("11. The current folder.").

I thought it might fail for you if the tests are run so that the current directory is something else than where liblzma.dll is. CMake's docs say that `WORKING_DIRECTORY` defaults to `CMAKE_CURRENT_BINARY_DIR` though, so maybe this guess is wrong.

Relying on the working directory isn't ideal anyway because if someone happened to have liblzma.dll in the system or Windows directory, then the DLL search order would pick that in step 7 or 8. (I'm assuming that safe DLL search mode is enabled.)

Seems that it's best to have the DLL and EXE files in the same directory. I don't like the idea of making a copy of liblzma.dll. I rather put the test executables in the top-level build directory. I pushed [cmake_tests_bin](https://github.com/tukaani-project/xz/compare/master...cmake_tests_bin). Does it fix the issue for you?

BraveChicken1:
I am using the plain MSVC toolchain on Windows. MSVC puts the binaries in a `Release` or `Debug` directory (depending on the build type). I think this could have affected the DLL search order.

The fix that you pushed on the `cmake_tests_bin` branch does indeed fix my issues. I agree that copying liblzma.dll is not the nicest solution, however my thought was that having the xz binaries and test binaries in the same directory also wouldn't be ideal.

Larhzu:
Thanks for testing! I committed it to master.

I also changed msvc.yml to use `-DBUILD_SHARED_LIBS=ON`. I tried it first without the fix, and the CI hanged when trying to run the tests. After the fix the CI works.

It's slightly messy that the test binaries are the top level directory, but it's better than copying the DLL.

PR #240, opened 2026-08-12
xz: Avoid integer overflow in --list totals

xordanblu:
`xz --list` reads the per-file sizes from the Index without decompressing Blocks. Each value is limited to `LZMA_VLI_MAX`, but `update_totals()` combines multiple files in `uint64_t` fields without checking the additions. Two maximum Uncompressed Size values followed by 2 therefore wrap the robot total from `2^64` to zero.

Minimal 136-byte reproducer set:

```sh
printf %s '/Td6WFoAAAD/EtlBAAAAAAAAAAAAAQX//////////3/1XeO8DdNWNwMAAAAAAFla' | base64 -d > max1.xz
cp max1.xz max2.xz
printf %s '/Td6WFoAAAD/EtlBAAAAAAAAAAAAAQUCQiD/swZynnoBAAAAAABZWg==' | base64 -d > two.xz
xz --robot --list max1.xz max2.xz two.xz
```

Current `master` (`c8b8ab2ef1eb0a0217ad2027d7f5d242ceb944d3`) ends with:

```text
totals 3 3 136 0 --- None 0 3
```

(The actual output is tab-separated.) A script using the documented robot total as a size or quota gate can consequently treat attacker-controlled metadata describing `2^64` bytes as zero.

This patch checks every numeric accumulation before modifying the totals. On overflow, xz exits with status 1, reports `The totals exceed the supported 64-bit range`, and omits the misleading partial `totals` line while preserving the individual file rows.

Validation performed:

- Release build completed with CMake/Ninja.
- All 20 CTest tests passed.
- `git diff --check` passed.
- Current open and closed issues/PRs were searched for the same `list.c` totals overflow; no matching report was found.

Discovery note: a time-boxed local AI-assisted review pointed to the existing integer-overflow TODO. I independently constructed and validated the binary reproducer against the exact upstream commit, checked for duplicates, implemented the patch, and ran the full test suite.

### DIFF
--- a/src/xz/list.c
+++ b/src/xz/list.c
@@ -213,6 +213,11 @@ static struct {
 	bool all_have_sizes;
 } totals = { 0, 0, 0, 0, 0, 0, 0, 0, 50000002, true };
 
+/// If calculating the totals overflows, don't print a partial totals line.
+/// In particular, --robot output may be used by scripts that expect the
+/// numeric fields to be trustworthy.
+static bool totals_overflow = false;
+
 
 /// Initialize colon_strs_fw[].
 static void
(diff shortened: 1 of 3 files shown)


Larhzu:
Thanks! It's true that it might matter with `--robot`.

I think only the uncompressed size can overflow easily. The others require very huge files. It's best to check for overflows in all of them still.

Your current patch calls `message_error` as soon as an overflow is detected, then the rest of the files are processed. That is, information about more files may be printed after the error message, so the error message might be difficult to notice. Options:

1. Delay `message_error` until all files have been processed.

2. Use `message_fatal` instead of `message_error`, and thus avoid the need for `totals_overflow`.

Considering that the overflow situation is rare and most likely happens with crafted/invalid files, I think the option 2 would be good enough. Scripts that use the robot mode hopefully check the exit status and ignore the output on any error. So in robot mode it likely doesn't provide much value to keep processing more files after an overflow has been detected.

What do you think?

xordanblu:
Thanks! Option 2 makes sense to me.

I changed the overflow path to `message_fatal`, removed `totals_overflow`,
and kept the checks for every accumulated numeric field.

I also strengthened the regression with a fourth input so it distinguishes
the immediate-fatal behavior. On the previous PR head, the focused test fails
with:

```
xz -l continued processing files after a totals overflow
```

With the updated commit, it stops after the third file, emits no `totals` row,
and the test passes. The Release build and full CTest suite pass 20/20, and
`git diff --check` passes too.

Larhzu:
I committed the list.c change with a different error message, an extra comment, and my own commit message. I added a test with a different method.

Thanks!

PR #243, opened 2026-09-03
liblzma: mt dec: Protect progress_in update with mutex

Trithem90:
`worker_decoder()` moves completed Block input progress into
`coder->progress_in` while holding `coder->mutex`.

At the same time, the main thread can decode the next Block Header
and update the same counter without holding the mutex. This allows
the two read-modify-write operations to race.

ThreadSanitizer detected concurrent writes between `worker_decoder()`
and `stream_decode_mt()` while decoding a valid multi-Block `.xz`
stream.

The race also produced an observable lost progress update. In one
reproduction after successful decoding:

    strm.total_in:        1048944
    lzma_get_progress():   917852
    difference:           131092 bytes

Protect the Block Header `progress_in` update with `coder->mutex`,
consistently with `worker_decoder()` and `get_progress()`.

Verification:

- Full CMake build succeeds.
- 20/20 upstream tests pass.
- Before the change, ThreadSanitizer reproduced the `progress_in`
  race together with the progress mismatch.
- After the change, 20 repeated TSan runs produced:
  - 0 ThreadSanitizer races
  - 0 non-monotonic progress results
  - 0 final progress mismatches

AI disclosure: AI tools assisted with drafting the test harnesses and
this description. I personally ran and reviewed the reproductions,
ThreadSanitizer output, functional progress mismatch, patch, and
post-fix results.


Larhzu:
Thanks! This was a good find. I edited the comment and commit message a bit, and committed the fix. It will be in 5.8.4, and I will backport it to v5.6 and v5.4 branches too.

> AI disclosure: AI tools assisted with drafting the test harnesses and
this description. I personally ran and reviewed the reproductions,
ThreadSanitizer output, functional progress mismatch, patch, and
post-fix results.

Thanks for this, especially that you verified things yourself before making the PR. :-)

PR #246, opened 2026-09-13
liblzma: Add DEC Alpha BCJ filter

mattst88:
Adds a BCJ filter for DEC Alpha. It converts BSR displacements and the `ldah $gp,X($ra|$pv)` / `lda $gp,Y($gp)` pairs that load the global pointer. These pairs follow every function entry and every call, and they account for most of the gain. BR displacements are converted relative to the most recent pair that uses `$pv`, which is at the start of the current function. On 837 executables and shared libraries (214 MB) from a Gentoo/alpha system, the filter improves compression by 10.08% over LZMA2 alone, and every file gets smaller. For comparison, the x86 filter gets 3.94% on a similar x86-64 system.

The series also adds `xz --alpha`, CI coverage, test files, and `test_bcj_alpha`.

## Needs approval

- **Filter ID 0x0C** is self-assigned. The .xz spec bump to 1.3.0 (d9b7abe6) is a proposal.
- **Version numbers:** the raw API is marked `\since 5.9.1alpha` under a new `XZ_5.10` symbol version, and `xz -lvv` reports that the file needs 5.10.0. All three assume the filter ships in 5.10.

## Needs review

- **Matching rules**, which can't change once released:
  - The `ldah` base register must be `$ra` or `$pv`.
  - The `lda` must be within four instructions, and the first match wins.
  - Instructions between the two halves of a pair are not converted.
  - BR is relative to the position of the last pair whose `ldah` uses `$pv`, or to 0 if there hasn't been one. This makes the filter keep 4 bytes of state across calls, like the x86 filter. The raw API functions start from 0 on every call.
  - Conditional branches are not converted (it was a net loss).
- **Compilers:** only checked against GCC output. No Tru64 or DEC C binaries were available.
- **Object files:** unlinked `.o` files get bigger, more than with the x86 filter: 0.98% over 937 objects from an Alpha defconfig kernel build, worst file 9.00%. Converting BR accounts for 0.34 points of that. The man page's existing note about object files covers this.
- **Raw API:** `lzma_bcj_alpha_encode()` and `lzma_bcj_alpha_decode()` are optional. They are in the last commit, with their test, so it can be dropped cleanly.


src/liblzma/simple/alpha.c
```diff
@@ -0,0 +1,156 @@
...
+
+		const uint32_t new_low = value & 0xFFFF;
+		const uint32_t new_high
+			= (value - alpha_sign_extend_16(new_low)) >> 16;
+
+		write32le(buffer + i, (instr & 0xFFFF0000)
+				| (new_high & 0xFFFF));
+		write32le(buffer + i + k, (low & 0xFFFF0000) | new_low);
+
+		// Skip the "lda" so that no other "ldah" can claim it.
```
ssvb: You can probably rewrite the code to just remember the last seen occurrence of `ldah`. Then, upon encountering `lda`, check how far the last `ldah` was and process the pair depending on the distance.

When handling the sequence of 3 back to back instructions **`ldah` (1), `ldah` (2), `lda`**, your current code is going to use the **`ldah` (1) + `lda`** pair. Would the **`ldah` (2) + `lda`** pair be a more correct choice?

ssvb:
> the filter improves compression by 9.62% over LZMA2 alone

Did you use `--lzma2=preset=6,lc=2,lp=2,pb=2` for compressing DEC Alpha code? The default `lc=3,lp=0,pb=2` LZMA parameters are not exactly optimal for the fixed size 4-byte instructions. This is some elaborate file specific tuning, but so is the BCJ filter too.

> BR is not converted (it was a net loss)

I wonder if anything can be done about this? I guess, these are very short distance conditional branches, which are local to a typically relatively small function. Replacing short distance relative branches with fat global absolute addresses inside of this function is predictably a loss, because no other code from the other functions is going to branch to these addresses. However, considering that the `ldah+lda` pairs are supposed to be used in the function prologues, can we try to use the address of the last encountered `ldah+lda` pair as the base address for the function local absolute address specifically for `BR` handling?

mattst88:
Hey @ssvb, long time no see! :)

> Did you use `--lzma2=preset=6,lc=2,lp=2,pb=2` for compressing DEC Alpha code?

No, both sides of the comparison used the defaults. I've now measured it on the same corpus (837 objects, 214 MB, whole file, preset 6):

| | default | lc=2,lp=2,pb=2 | lc=0,lp=2,pb=2 |
|---|---|---|---|
| no filter, vs. default | 0 | -0.88% | -1.06% |
| filter, vs. no filter with the same options | -9.62% | -9.69% | -9.70% |

The tuned options are worth about 1% on their own and don't change what the filter gains, so the two add up. The man page part of this PR already recommends `pb=2,lp=2,lc=2` for Alpha.

> can we try to use the address of the last encountered `ldah+lda` pair as the base address for the function local absolute address specifically for `BR` handling?

That works, but only if the base comes from the `ldah $gp,X($pv)` pairs, which are at function entry. Most pairs are `ldah $gp,X($ra)` reloads after a `jsr`, in the middle of a function, and using those as the base loses about half of the gain.

| variant | corpus | files improved vs. no BR | worst file vs. no BR | vmlinux |
|---|---|---|---|---|
| BSR + pairs (current PR) | -9.62% | | | -5.92% |
| + BR relative to last $pv pair | -10.08% | 829/837 | +0.27% | -6.10% |
| + BR relative to last pair of either kind | -9.81% | 810/837 | +0.53% | -5.86% |
| + BR absolute | -9.20% | 703/837 | +3.17% | -5.27% |
| + conditional branches relative to last pair | -8.61% | 64/837 | +3.79% | -2.16% |

One detail: `BR` on Alpha is the unconditional branch, nearly always with Ra = $31. The conditional branches are opcodes 0x38-0x3F, and treating them the same way makes compression worse. I also tried resetting the base after `ret` or at `unop` alignment padding, to handle leaf functions that don't set up $gp. Both were worse than the $pv base alone.

Compared to no filter, every file still gets smaller (median -6.26%, worst -1.47%).

I've updated the PR with the $pv variant. The filter now keeps 4 bytes of state across calls, the position of the last $pv pair, similar to the x86 filter. Decoding still works because the pairs are found using only instruction fields the filter doesn't change. The test file generator and `test_bcj_alpha` now cover BR as well.

ssvb:
It's interesting that `lc=0,lp=2,pb=2` works slightly better in your tests. If this is consistently reproducible, then maybe recommend this configuration in the man page? Also maybe try the whole space of all possible lc/lp/pb configurations together with the BCJ filter to see if anything works even better? That is unless you have already done that and `lc=0,lp=2,pb=2` turned out to be the best.

> The conditional branches are opcodes 0x38-0x3F, and treating them the same way makes compression worse.

Regarding the conditional branches. Maybe try to keep relative addressing, but encode the jump offset not using two's complement representation, but store the sign in the lowest bit? Something like this:
```C
uint32_t twoscomplement_to_lowbitsign(int32_t v) {
    return (v >= 0) ? (((uint32_t)v) << 1) : (((~(uint32_t)v) << 1) + 1);
}

int32_t lowbitsign_to_twoscomplement(uint32_t v) {
    return (v & 1) ? ~(v >> 1) : (v >> 1);
}
```
So that `0x00000012` becomes `0x00000024` and `0xFFFFFFFF` (-1) becomes `0x00000001`. The rationale is that the upper bytes of short distance conditional branch instruction opcodes will be filled by zero bits regardless of the branch direction (resulting in more similarity in the instruction upper bytes across the whole program).

ssvb:
> It's interesting that `lc=0,lp=2,pb=2` works slightly better

Actually byteswapping each 32-bit instruction opcode to big endian might fix that. The `lc` parameter leverages dependency on high bits of the previous byte, and uses these bits as the context for prediction. If DEC Alpha instructions are stored as little endian, then the first byte is a random-ish displacement/immediate operand, which is useless for predicting what comes next. But if the instructions are stored as big endian, then the first byte encodes the instruction type, and the following bytes can be partially predicted based on it. At least UTF-16BE/UTF-32BE text typically compresses better than UTF-16LE/UF-32LE: https://github.com/tukaani-project/xz-java/issues/19#issuecomment-2689315878

ssvb:
> Compilers: only checked against GCC output. No Tru64 or DEC C binaries were available.

Out of curiosity, is there a working LLVM/Clang for DEC Alpha? I see that there's https://github.com/alphalinux-org/llvm-project repository in your github organization.

mattst88:
Thanks, I measured all three. The corpus has been re-extracted from the same system (836 objects, 210 MB, package updates since last time). Everything is raw LZMA2 on whole files at preset 6, and every variant is round-trip checked. With default options the current filter comes out at -10.13% here.

**All lc/lp/pb combinations** (75 of them, lc+lp <= 4, pb 0-4), total size vs. no filter with default options:

| | best | 2nd best | lc=2,lp=2,pb=2 | default |
|---|---|---|---|---|
| no filter | lc=0,lp=2,pb=2: -1.06% | lc=1,lp=2,pb=2: -1.05% | -0.87% | 0 |
| filter | lc=1,lp=2,pb=2: -11.23% | lc=0,lp=2,pb=2: -11.20% | -11.01% | -10.13% |

lp=2,pb=2 always wins, and lc=0 and lc=1 are tied. Per file, lc=0 beats lc=2 in 806 of 836 files. The kernel goes the other way, though (vmlinux: lc=2 -7.07%, lc=0 -6.99%). The difference is about 0.2%, and the man page sentence is shared with ARM64, so I'd rather leave it as is unless Lasse prefers otherwise.

**Byteswapping to big endian** makes things worse:

| | default | lc=0,lp=2,pb=2 | files improved | vmlinux |
|---|---|---|---|---|
| filter + byteswap, vs. filter | +1.94% | +2.02% | 8-11/836 | -3.33% vs. -6.10% |

Without the filter it's +1.26%. lc=0 is already the best setting in both byte orders, so literal context doesn't seem to matter much for this data. That would explain why the UTF-16/UTF-32 effect doesn't carry over.

**Sign in the low bit** for the conditional branch displacement, relative to the current filter with the same options:

| variant | default | lc=0,lp=2,pb=2 | files improved | worst file | vmlinux |
|---|---|---|---|---|---|
| + Bcc (0x38-0x3F) sign-in-low-bit | +0.25% | +0.18% | 262/836 | +1.09% | -4.58% vs. -6.10% |
| JSR hint (14-bit) sign-in-low-bit | +0.81% | +0.79% | 0/836 | +4.44% | -5.30% |
| BR sign-in-low-bit instead of relative to the last $pv pair | +0.68% | +0.67% | 6/836 | +2.60% | -5.82% |

It's better than making Bcc function-relative (+1.32%), but it still loses to leaving the field alone, so I haven't changed the filter.

> Out of curiosity, is there a working LLVM/Clang for DEC Alpha? I see that there's https://github.com/alphalinux-org/llvm-project repository in your github organization.

Yes, still a work-in-progress, but it builds (and boots) Linux and can `emerge @world` in a Gentoo stage3!

ssvb:
> Thanks, I measured all three.

Thanks for the detailed measurements. I apologize for kinda throwing random ideas at the wall to see what sticks.

> Yes, still a work-in-progress, but it builds (and boots) Linux and can <code>emerge @world</code> in a Gentoo stage3!

So the binaries compiled by this experimental Clang can be, in principle, tested too? I still wonder why the results are so different between vmlinux and that 210MB collection of executables. Do they maybe use different optimization options (`-Os` vs. `-O2` vs. `-O3`)?

For the sake of completeness, the "6e" preset (--extreme) could be tested too. It's much slower, but useful if the users can afford longer compression times.

And I wonder if the `--alpha` command line option could automatically override the default to `lp=2,pb=2` as a minor quality of life improvement for the users? Maybe for the ARM64 BCJ filter too. That's an extra ~1% compression rate boost left on the table if the users forget to tweak these parameters. This decision is entirely up to Lasse.

ssvb:
> Byteswapping to big endian makes things worse

Are you sure about this? I tried to apply the following simple test patch on top of your pull request (it does conversion to big endian as a slow separate pass over the data buffer) and see very measurable improvements using the `alpha-linux-gnu-gcc` crosscompiler packaged in Debian to compile https://sqlite.org/2026/sqlite-amalgamation-3530400.zip sources:

```diff
diff --git a/src/liblzma/simple/alpha.c b/src/liblzma/simple/alpha.c
index 503bafa8..954b407f 100644
--- a/src/liblzma/simple/alpha.c
+++ b/src/liblzma/simple/alpha.c
@@ -64,6 +64,10 @@ alpha_code(void *simple_ptr, uint32_t now_pos, bool is_encoder,
 	lzma_simple_alpha *simple = simple_ptr;
 	uint32_t func_pos = simple->func_pos;
 
+	if (!is_encoder) // Byteswap from BE to LE in the decoder
+		for (size_t j = 0; j < size; j += 4)
+			write32le(buffer + j, read32be(buffer + j));
+
 	const size_t limit = size - ALPHA_LOOKAHEAD;
 	size_t i = 0;
 
@@ -144,6 +148,14 @@ alpha_code(void *simple_ptr, uint32_t now_pos, bool is_encoder,
 		i += k + 4;
 	}
 
+	if (!is_encoder) // Byteswap the unused part back from LE to BE in the decoder
+		for (size_t j = i; j < size; j += 4)
+			write32be(buffer + j, read32le(buffer + j));
+
+	if (is_encoder) // Byteswap from LE to BE in the encoder
+		for (size_t j = 0; j < i; j += 4)
+			write32be(buffer + j, read32le(buffer + j));
+
 	simple->func_pos = func_pos;
 	return i;
 }

```

mattst88:
> Are you sure about this? I tried to apply the following simple test patch [...] sqlite-amalgamation

I think it depends on whether the code is linked. I cross-built the same sqlite amalgamation with GCC 16.1 at -O0/-O1/-Os/-O2/-O3 (with and without -mcpu=ev67) and compared filter+byteswap against the filter alone, preset 6:

| | byteswap vs. filter | filter vs. no filter |
|---|---|---|
| `sqlite3.o` (compiled with -c, not linked) | -0.8% to -3.5% in 9 of 10 builds | +2.8% to +12% |
| `sqlite3` linked (PIE) | +0.1% to +2.2% in all 10 builds | -5.7% to -8.7% |

The -O2 build linked non-PIE, static, and stripped gives +1.26%, +2.37%, and +0.61% for byteswap.

In a `.o` file, most `ldah`/`lda` displacements are still zero, waiting for relocations (57% of them in `sqlite3.o`). The filter makes those files larger anyway, which the man page's existing note about object files covers. On linked binaries byteswapping loses every time.

Also, the test patch doesn't round-trip: `xz --alpha` followed by `xz -d` doesn't give back the original sqlite3 binary. It swaps all `size` bytes, including the unfiltered tail at the end of the buffer, which the simple coder passes in again on the next call, so those words get swapped twice. On the -O2 sqlite3 it compresses to 689952 bytes, vs. 687771 for a byteswap done correctly.

> So the binaries compiled by this experimental Clang can be, in principle, tested too?

Yes. I took the 1401 files built by Clang/lld on a Gentoo/alpha system rebuilt with Clang (922 MB), whole file, raw LZMA2:

| | preset 6 | 6, lc=0,lp=2,pb=2 | 6e |
|---|---|---|---|
| filter (this PR) | -12.57% (median -8.24%, 1387/1401 improved) | -12.57% | -12.80% |
| filter + byteswap | -10.50% | -10.68% | -10.78% |

That set includes shared libraries and LLVM itself, which the earlier GCC set didn't: the earlier set turned out to be only `/usr/bin`, not "executables and shared libraries" as I wrote in the PR description. Comparing just `/usr/bin` is fairer:

| `/usr/bin` only, preset 6 | files | size | filter vs. no filter |
|---|---|---|---|
| GCC | 836 | 210 MB | -10.13% (median -6.24%) |
| Clang | 708 | 209 MB | -12.32% (median -8.36%) |

Shared libraries built with Clang (`/usr/lib` without LLVM, 529 files, 233 MB) get -10.53%.

Clang also uses a gp setup that GCC hardly uses, `br $gp,.+4; ldah $gp,X($gp); lda $gp,Y($gp)`, at 0.43% of instruction words. The PR doesn't catch it yet. Accepting `$gp` as the `ldah` base register gives -13.39% on the Clang corpus. Also using that pair as the BR base, like the `$pv` pair, gives -13.67% (worst file +0.38% vs. the current filter). It changes GCC output by -0.01% and vmlinux not at all. Pair detection still only reads fields the filter doesn't change, so it stays reversible.

> I still wonder why the results are so different between vmlinux and that 210MB collection of executables. Do they maybe use different optimization options

Both are -O2. The difference is BSR. Userspace BSRs mostly come from linker relaxation, which rewrites `ldq $pv,...($gp); jsr` into `bsr` when the target is in range. `sqlite3.o` has no BSRs at all, the linked `sqlite3` has 3.37%, and linking with `-Wl,--no-relax` brings it back to zero. The kernel links with `-static -N` and `-relax` is commented out in `arch/alpha/Makefile`, so vmlinux has 24 BSRs in 1.75M instructions. The gpdisp pair density is about the same in both.

> the "6e" preset (--extreme) could be tested too

It changes very little. Filter vs. no filter at 6e is -10.14% on the GCC corpus (-10.13% at 6), -5.94% on vmlinux (-6.10%), and -12.80% on the Clang corpus (-12.57%).

> And I wonder if the `--alpha` command line option could automatically override the default to `lp=2,pb=2`

That's worth about another 1% (-1.06% on GCC output, -0.84% on Clang output), on top of the filter's gain.

ssvb:
> the test patch doesn't round-trip: xz --alpha followed by xz -d doesn't give back the original sqlite3 binary.

Yes, my initially posted patch was indeed incorrect, but I quickly-sh edited that comment to fix it after posting. You still managed to grab the initial broken version in that short window.

> In a .o file, most ldah/lda displacements are still zero, waiting for relocations (57% of them in sqlite3.o). The filter makes those files larger anyway, which the man page's existing note about object files covers. On linked binaries byteswapping loses every time.

There are things other than the cpu instructions in executable files. Such as data sections, unstripped debugging symbols, etc. Byteswapping this other data seriously ruins its compressibility. But it's possible to extract the code section from an executable or a shared library via:
```sh
alpha-linux-gnu-objcopy --only-section=.text -O binary sqlite-executable sqlite-only-code-section.bin
```

Of course, this remains non-practical until *xz* is further improved to easily (and preferably automatically) enable this filter only for that certain parts of a file, where applying it is really beneficial. Even now there's the ` --block-list=` command line option, but it is inconvenient.

